PR #9, opened 2019-08-05
Prevent indefinite hangs when writing to the node monitoring UNIX socket

julesdesmit:
This commit should prevent a possible deadlock of the logrus instance, which can be caused if the node monitoring component hangs for any reason and does not read incoming messages. When this keeps going, the buffer can fill up and subsequent writes will block until this buffer is cleared.

Because we write to the node monitoring component with a logrus hook, the associated mutex is acquired and not released for as long as the write operation hangs.

### DIFF
--- a/pkg/eventmon/logger/logger.go
+++ b/pkg/eventmon/logger/logger.go
@@ -30,6 +30,7 @@ type (
 		p                 wire.EventPublisher
 		entry             *log.Entry
 		acceptedBlockChan <-chan block.Block
+		EntryChan         chan *log.Entry
 		listener          *wire.TopicListener
 	}
 )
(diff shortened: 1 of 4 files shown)


ggoranov:
The solution looks good to me. Just one question:

As you pointed out:
`Because we write to the node monitoring component with a logrus hook, the associated mutex is acquired and not released for as long as the write operation hangs.`

As we've got now 3 seconds write-deadline, wouldn't the node hang for 3 seconds repeatedly next time when the monitoring component hangs again?

@julesdesmit

julesdesmit:
@ggoranov You're absolutely right - that's a good point. It might be best to also sever the connection and stop logging if we miss the write deadline. I will add it.

autholykos:
@julesdesmit I think we should modify the hook to actually separate the streaming part from the log part by buffering the logs (and start dropping them if the connection is really slow or severed). This way, we can decouple the mutex locking due to the log transmission from the actual logging and the node would be unaffected by any inefficiency of the hook

julesdesmit:
@autholykos Even better, actually - I'll implement it.

PR #10, opened 2019-08-07
GraphQL support

ggoranov:


launch/testnet/server.go
```diff
@@ -61,6 +63,18 @@ func Setup() *Server {
...
 
+	// Instantiate GraphQL server
+	if cfg.Get().Gql.Enabled {
+		gqlServer, err := gql.NewHTTPServer(eventBus, rpcBus)
+		if err != nil {
+			log.Errorf("graphQL server error: %s", err.Error())
+		}
+
+		if err := gqlServer.Start(); err != nil {
+			log.Errorf("graphQL server error: %s", err.Error())
```
julesdesmit: Would it be better to have distinct error messages for these two function calls? Instead of the current one, we could do something like: `error creating graphQL server` and `error starting graphQL server`, to make it more clear where the error is coming from.

pkg/gql/query/blocks.go
```diff
@@ -0,0 +1,208 @@
...
+
+	blocks := make([]*block.Block, 0)
+	err := b.db.View(func(t database.Transaction) error {
+		for _, v := range hashes {
+			encodedHash, ok := v.(string)
+			if !ok {
+				continue
+			}
+
+			hash, err := base64.StdEncoding.DecodeString(encodedHash)
```
julesdesmit: Maybe it would be better to take in hex strings, instead of base64, as hex strings are the common encoding for most tx/block related information.

autholykos: Yes. Maybe it would make sense to use a custom type for `hash` with a predefined `String()` method that Hex encodes the `[]byte`?

launch/testnet/server.go
ggoranov: Good catch!

pkg/gql/query/blocks.go
ggoranov: > Maybe it would be better to take in hex strings, instead of base64, as hex strings are the common encoding for most tx/block related information.

Indeed! I chose Base64 as it gives a shorter representation than Hex. But as we already have the wallet working with Hex encoding, better to be aligned with that one for now.

ggoranov: > Yes. Maybe it would make sense to use a custom type for hash with a predefined String() method that Hex encodes the []byte?

Better to handle this with another task where core packages can be updated accordingly too.

julesdesmit: Agreed, feel free to open an issue.

PR #14, opened 2019-08-12
Eventbus handler

### LINKED ISSUE #12: Refactor event bus handlers
With the addition of the new ring buffer, the Handler interface now contains behaviour which is only needed for one specific implementation. This can probably be refactored to be a bit more accomodating.

autholykos:
This fixes #12 and introduces some side cosmetic changes (mainly comments and DRY logs)


autholykos:
`wire.Handler` is now a cleaner interface, consistent among the different methodologies (callback, channel and ring-buffer). I also eliminated some unnecessary one-liner methods from `wire.EventBus`

PR #18, opened 2019-08-13
Remove the need to sync the wallet manually

### LINKED ISSUE #15: Call sync command automatically when doing anything wallet-related
As pointed out, needing to sync manually any time a wallet operation is performed is quite cumbersome. This should be integrated into the wallet functions themselves, instead.

julesdesmit:
Fixes #15.

### DIFF
--- a/pkg/cli/walletcommands.go
+++ b/pkg/cli/walletcommands.go
@@ -94,7 +94,6 @@ func loadWalletCMD(args []string, publisher wire.EventBroker, rpcBus *wire.RPCBu
 }
 
 func loadWallet(password string) (*wallet.Wallet, error) {
-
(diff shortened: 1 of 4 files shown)


pkg/cli/walletcommands.go
```diff
@@ -142,6 +141,12 @@ func transferCMD(args []string, publisher wire.EventBroker, rpcBus *wire.RPCBus)
 		return
 	}
 
+	// Sync wallet
+	if err := syncWallet(); err != nil {
+		fmt.Fprintf(os.Stdout, "%v", err)
```
autholykos: why the error is printed and not returned? The function clearly returns `(*Wallet, error)`

PR #23, opened 2019-08-13
Clarify purpose and behaviour of `startprovisioner` and `startblockgenerator`

### LINKED ISSUE #7: Clarify the effects of `startprovisioner`
It is not so intuitive that you can execute startprovisioner without staking. According to the documentation it is necessary to stake.

```"stake": Stake a given amount of DUSK, to allow participation as a provisioner in consensus. Make sure to use the sync command first.```

What is actually happening when you run startprovisioner without staking?

Is it a deliberate choice to let users sync manually instead of syncing automatically? What is the risk of staking without syncing?

### PR DESCRIPTION
**Pull request by julesdesmit**
Fixes #7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,10 @@ To claim Testnet DUSK (tDUSK), the user is required to make a Twitter post conta
 ### How to transfer tDUSK?
 Open the Command Line Interface (CLI) and type `transfer [amount] [address] [password]`, where `[amount]` stands for the amount of tDUSK the user is willing to bid (`0 < amount <= balance`), `[address]` stands for the recipient address, and `[password]` stands for the wallet password. 
 ### How to become a Provisioner?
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ggoranov** [comment]
@julesdesmit @autholykos Thinking about the `startprovisioner/startblockgenerator` cli commands, I was wondering if we need them as cli commands at all? Note that we don't plan to provide `stop_provisioner/stop_blockgenerator` commands, do we?. In addition, as per the latest changes `startblockgenerator` does not need input anymore.

  Instead, we can have both of them enabled by a config. For instance:

```toml

[consensus]
  provisionerEnabled = true
  blockGeneratorEnabled = true
```

julesdesmit:
@ggoranov We actually already have an issue open for this: #27. Goal being to automate the startup of consensus, and stake/bid management by the node so the operator only needs to turn it on and deposit some DUSK.

It's also right that those changes would make this PR obsolete, as we could get rid of a lot of the README stuff, including this.

ggoranov:
> @ggoranov We actually already have an issue open for this: #27. Goal being to automate the startup of consensus, and stake/bid management by the node so the operator only needs to turn it on and deposit some DUSK.
> 
> It's also right that those changes would make this PR obsolete, as we could get rid of a lot of the README stuff, including this.

Indeed. #27 would handle this gracefully. Let's close this MR then.

PR #37, opened 2019-08-20
Update chain to sync committee with consensus committees on launch

### LINKED ISSUE #36: Chain committee is one round behind consensus committee during launch
Since a `committee.Store` is initialized on the chain tip height, and a consensus round is always one ahead of the tip, it can cause a discrepancy between the chain and the consensus committees.

For example, if the chain tip is at block 2000, then the round is 2001. Upon startup, consensus components run `RemoveExpiredProvisioners` with round 2001, so their committee is then one step ahead of the chain, who only calls this function when blocks are accepted.

As a result, this can cause the node to get stuck, since a forged block will have an invalid certificate according to the chain, even though it is correct from a consensus point of view.

julesdesmit:
Fixes #36

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -74,6 +74,9 @@ func New(eventBus *wire.EventBus, rpcBus *wire.RPCBus, c committee.Foldable) (*C
 		certificateChan: certificateChan,
 	}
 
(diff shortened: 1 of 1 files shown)


ggoranov:
LGTM

PR #45, opened 2019-08-21
Retry random key generation, in case of EOF

### LINKED ISSUE #44: Fix intermittent failure of BLS keypair generation
When calling `user.NewRandKeys` or `user.NewKeysFromBytes`, if the `rand.Reader`, or any other seed, runs out of bytes before being able to generate a BLS keypair, it returns an EOF error. This can be easily mitigated by adding some logic to retry with a new seed in this case. Additionally, this should take care of an intermittently failing test.

julesdesmit:
Fixes #44

### DIFF
--- a/pkg/core/consensus/user/keys.go
+++ b/pkg/core/consensus/user/keys.go
@@ -21,9 +21,14 @@ type Keys struct {
 // NewRandKeys will generate and return new bls and ed25519
 // keys to be used in consensus
 func NewRandKeys() (Keys, error) {
-	keys, err := NewKeysFromReader(nil)
-	if err == io.EOF {
-		return NewRandKeys()
+	var keys Keys
+	var err error
+	for {
+		keys, err = NewKeysFromReader(nil)
+		if err == io.EOF {
+			continue
+		}
+		break
 	}
 
 	return keys, err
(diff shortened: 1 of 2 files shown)


pkg/core/consensus/user/keys.go
```diff
@@ -21,7 +21,12 @@ type Keys struct {
 // NewRandKeys will generate and return new bls and ed25519
 // keys to be used in consensus
 func NewRandKeys() (Keys, error) {
-	return NewKeysFromReader(nil)
+	keys, err := NewKeysFromReader(nil)
+	if err == io.EOF {
+		return NewRandKeys()
```
ggoranov: Is there a reason for using recursion instead of a simple loop here?

julesdesmit: It's more concise, I think. Other than that, no reason.

autholykos: In general I would avoid recursion. In this case though, recursion helps the conciseness and I find it more elegant than a 'for' with a nested if. At the end of the day the recursion is a rare circumstance while a 'for' would suggest that an EOF is expected to happen less infrequently

ggoranov: My point was this would grow the stack indefinitely. If a `for` was used the compiler would choose to simply inline `NewRandKeys`

autholykos: Only if the recursion actually happens, which is a rare occurrence. However, point taken. Feel free to revert to a for loop

PR #67, opened 2019-09-03
Automate the startup of consensus, and the sending and maintaining of bids and stakes

### LINKED ISSUE #27: Automate startup of node, with regards to consensus
With both the provisioner component and the generation component not being dependent on a specific hash, and with both now able to update their values when a tx expires, we can make the node even more easy to use by automating the syncing, staking, bidding and starting of consensus components without any need for user input.

Some handy added features would be specifying a `--wallet` flag to start the node without consensus running, and adding some commands that can adjust default locktimes and values for automated transactions.

This enhancement should remove the need for node operators to consistently check if their transactions have expired and from sending new ones in due time by themselves, which is much less of a drag. This replaces #13, since a user can no longer accidentally start consensus components on the wrong round this way.

### LINKED ISSUE #54: Consensus is not initialized when running the network from genesis block
When starting a local network from scratch (for testing purposes) it seems no nodes are running the consensus.

According to the current implementation, `msg.InitializationTopic` is published in case of topics.AcceptedBlock event only which is totally fine for updating the consensus module with the chain tip. 

However, in case of starting the network from genesis block, wouldn't the network get stuck as no node is able to generate block at height 1?

@julesdesmit  The below patch workarounds this but need your expertise here, if I'm missing something?
```
+++ b/pkg/core/consensus/initiator.go
@@ -35,6 +35,13 @@ func GetStartingRound(eventBroker wire.EventBroker, db database.DB, keys user.Ke
        // Start listening for accepted blocks, regardless of if we found stakes or not
        acceptedBlockChan, listener := InitAcceptedBlockUpdate(eventBroker)
 
+       if currentHeight == 0 {
+               roundBytes := make([]byte, 8)
+               binary.LittleEndian.PutUint64(roundBytes, 1)
+               eventBroker.Publish(msg.InitializationTopic, bytes.NewBuffer(roundBytes))
+               return nil
+       }
+

```

### LINKED ISSUE #74: Consider exposing wallet commands as an API
Currently, the entire wallet interaction goes through CLI commands. A client application is not capable of creating a transaction, getting balance and so on. 
An example: To enable test-harness, rpc package is calling `cli.SendBidTx` directly as a temporary patch.

Proposal:

Wallet commands can be moved to a separate package so that both CLI and RPC  packages can use them via RPCBus.

Other option could be to deprecated CLI entirely and build RPC API only for interaction with the wallet.

Let me know your thoughts on that.
@julesdesmit @autholykos

julesdesmit:
Fixes #27, fixes #54, and fixes #74


pkg/core/consensus/initiator/initiator.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func startProvisioner(eventBroker wire.EventBroker, rpcBus *wire.RPCBus, w *wallet.Wallet) {
+	// Setting up the consensus factory
+	f := factory.New(eventBroker, rpcBus, config.ConsensusTimeOut, w.ConsensusKeys())
+	f.StartConsensus()
+
+	// Get starting round
+	blsPubKey := w.ConsensusKeys().BLSPubKeyBytes
+	startingRound := getStartingRound(blsPubKey, eventBroker)
```
ggoranov: With reference to #54, is there a specific reason we use "getStartingRound" (the one that waits for acceptedBlock event) to fetch the current chain tip, instead of fetching the tip from the `db.FetchCurrentHeight`?

ggoranov: Note that `blsPubKey`  is not anymore used with latest changes.

julesdesmit: Correct, I just ran into this while testing. Currently added a `GetLastBlock` call with the RPCBus to get our height. If it is 0, we start without waiting as per #54. One more thing I am currently remembering that wasn't yet finished, is for the node to wait to be synced to the tip. This is part of the reason as well why we use `getStartingRound`, the other being, that we could have a slight chance of mis-timing our entry and missing a lot of crucial messages for agreement, getting the node stuck instantly.

julesdesmit: Also, correct about `blsPubKey`. will be removed in the next commit, thank you for noticing!

pkg/core/consensus/initiator/initiator.go
```diff
@@ -0,0 +1,122 @@
...
+	roundBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(roundBytes, startingRound)
+	publisher.Publish(msg.InitializationTopic, bytes.NewBuffer(roundBytes))
+}
+
+func syncToTip(acceptedBlockChan <-chan block.Block, counter *chainsync.Counter) {
+	i := 0
+	for {
+		<-acceptedBlockChan
+		if counter.IsSyncing() {
```
ggoranov: Theoretically, once syncing procedure reaches 500 blocks or time-out, IsSyncing would return false and syncToTip could exit without reaching the network tip.

Would this affect the consensus?

pkg/cli/walletcommands.go
```diff
@@ -64,11 +67,39 @@ func createWalletCMD(args []string, publisher wire.EventBroker, rpcBus *wire.RPC
 	fmt.Fprintf(os.Stdout, "Wallet created successfully!\n")
 	fmt.Fprintf(os.Stdout, "Public Address: %s\n", pubAddr)
 
-	cliWallet = w
-	DBInstance = db
+	c.transactor = transactor.New(w, nil)
+
+	if !cfg.Get().General.WalletOnly {
+		go initiator.LaunchConsensus(c.eventBroker, c.rpcBus, w, c.counter)
```
ggoranov: nitpick: If LaunchConsensus only starts another go-routines, do we need `go` here?

pkg/core/transactor/transactor.go
```diff
@@ -0,0 +1,186 @@
...
+	// Sign tx
+	err = t.w.Sign(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	return tx, nil
+}
+
+func (t *Transactor) CreateStakeTx(amount, lockTime uint64) (transactions.Transaction, error) {
```
ggoranov: I think transactor.CreateStakeTx could be called concurrently (by cli and by maintainer processes). In addition, `w.Sign` does touch the wallet db. Wouldn't we need to handle this?

pkg/config/samples/default.dusk.toml
```diff
@@ -3,6 +3,7 @@
 # general node configs
 [general]
 network = "testnet"
+walletonly = false
```
ggoranov: Could we add some comments explaining what `walletonly` mode means for the node?

pkg/core/consensus/initiator/initiator.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/peer/processing/chainsync"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/encoding"
+	"github.com/dusk-network/dusk-blockchain/pkg/wallet"
+	log "github.com/sirupsen/logrus"
+)
+
+var l = log.WithField("process", "consensus initiator")
+
+func LaunchConsensus(eventBroker wire.EventBroker, rpcBus *wire.RPCBus, w *wallet.Wallet, counter *chainsync.Counter) {
```
ggoranov: As the task is about `Automate the startup of consensus`, don't we need to call LaunchConsensus at startup? I find it being called only within the CLI commands. (sorry probably I'm misunderstanding something here)

pkg/core/transactor/collector.go
```diff
@@ -0,0 +1,38 @@
...
+		transferChan chan transferInfo
+	}
+
+	transferInfo struct {
+		amount  uint64
+		address string
+	}
+)
+
+func initTransferCollector(subscriber wire.EventSubscriber) chan transferInfo {
```
ggoranov: Is `initTransferCollector` used anywhere, or it's here only as an example?

pkg/p2p/wire/topics/topics.go
```diff
@@ -40,6 +40,11 @@ const (
 	Agreement      Topic = "blockagreement"
 	StartConsensus Topic = "startconsensus"
 
+	// Transaction topics
```
ggoranov: Are these topics being used anywhere?

pkg/core/consensus/initiator/initiator.go
julesdesmit: > syncToTip could exit without reaching the network tip

It shouldn't, as it would need multiple blocks accepted in a row without triggering the sync counter before being able to exit.

```
     i++
     if i > 2 {
       break
     }
```

pkg/core/transactor/transactor.go
julesdesmit: Yeah, good point. There's a chance both of them would attempt to use the same inputs resulting in an accidental attempt to double-spend.

pkg/config/samples/default.dusk.toml
julesdesmit: Yep, will do

pkg/core/consensus/initiator/initiator.go
julesdesmit: It's really quite hard to start consensus if we have no wallet loaded. I think we can also automate that part but I'm not sure if we should take that out of the hands of the user either.

pkg/cli/walletcommands.go
```diff
@@ -211,14 +181,18 @@ func createFromSeedCMD(args []string, publisher wire.EventBroker, rpcBus *wire.R
...
-}
 
-func createFromSeed(seedBytes []byte, password string) (*wallet.Wallet, error) {
+	c.transactor = transactor.New(w, nil)
 
-	if DBInstance != nil {
-		DBInstance.Close()
+	if !cfg.Get().General.WalletOnly {
+		go initiator.LaunchConsensus(c.eventBroker, c.rpcBus, w, c.counter)
+		if err := c.launchMaintainer(w); err != nil {
```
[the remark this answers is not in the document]
julesdesmit: Yep, good point. Will change it

pkg/core/transactor/collector.go
julesdesmit: No, you're right. This is some leftover code that I forgot to remove, I modeled the transactor to work with channels at an earlier stage.

pkg/p2p/wire/topics/topics.go
julesdesmit: Same goes for this, leftover code. Will remove it

pkg/cli/walletcommands.go
julesdesmit: Good catch, I think I did it cause initially I was planning on syncing to tip before starting the other goroutines, though it is only needed for the provisioner component.

pkg/core/transactor/transactor.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Turn amount into a scalar
+	amountScalar := ristretto.Scalar{}
+	amountScalar.SetBigInt(big.NewInt(0).SetUint64(amount))
+
+	// Send amount to address
+	tx.AddOutput(key.PublicAddress(address), amountScalar)
```
ggoranov: Could we check the return error here?

julesdesmit: Yep, good catch

pkg/cli/commands.go
```diff
@@ -1,8 +1,6 @@
 package cli
 
 import (
-	"bytes"
-	"encoding/binary"
```
ggoranov: nitpick: stringToInt64, intToScalar, stringToScalar, generateDualKey in cli pkg seem are unused for now.

julesdesmit: Correct - removing them.

pkg/core/consensus/initiator/initiator.go
ggoranov: Could we load the wallet automatically?

   In case of a panic, the node should be capable of recovering without need of any manual actions like LoadWallet from cli. That said, I assume if wallet.dat and walletDB are existing, node can cope with reloading wallet and restarting the consensus without any manual action. Otherwise if a node is down and then restarted automatically, it will be out of network consensus until the node owner realizes it.

julesdesmit: That's a very good point, I hadn't considered it. I will move to implement it.

julesdesmit: Hmm, one thing that does come up is the fact that the node would need to know about the wallet password. Would there be a safe way to keep that somewhere?

pkg/core/consensus/initiator/initiator.go
ggoranov: Ah, good point. Makes sense.
One more question, so that means anytime the node is restarted and the wallet is loaded, it will always need to wait for at least 3 blocks accepted before initializing the consensus?

julesdesmit: It would need to wait for 3 accepted blocks in a row, where `IsSyncing()` returns false. Then we assume we are at the tip, and we start consensus.

pkg/core/consensus/initiator/initiator.go
ggoranov: Indeed.  Dunno safe way. We can decide on this later and just keep it in mind for now.

pkg/core/transactor/transactor.go
```diff
@@ -0,0 +1,186 @@
...
+	// Sign tx
+	err = t.w.Sign(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	return tx, nil
+}
+
+func (t *Transactor) CreateBidTx(amount, lockTime uint64) (transactions.Transaction, error) {
```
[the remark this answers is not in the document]
ggoranov: The suggestion here was more about making Transactor concurrent-safe.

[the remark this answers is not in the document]
julesdesmit: Ah I see. Apart from the walletDB though, there's not much that should be protected on this struct. However, after a quick look I believe it would be a good idea to lock while syncing, so I will implement it.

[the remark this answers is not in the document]
ggoranov: Up to you.

PR #68, opened 2019-09-04
Test harness

ggoranov:
More details in ./harness/README.md

Preconditions before merging:

1. Merge GraphQL support (tech-issue-48 branch)
2. Merge fix for https://github.com/dusk-network/dusk-blockchain/issues/27
3. Merge fix for https://github.com/dusk-network/dusk-blockchain/issues/54

For points 2 and 3 there are temporary patches within current branch


harness/engine/network.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	// Start voucher seeder
+	if len(seederExec) > 0 {
+		if err := n.start("", seederExec); err != nil {
+			return err
+		}
+	} else {
+		// If path not provided, then it's assumed that the seeder is already running
+		log.Warnf("Seeder path not provided.")
```
julesdesmit: Maybe it would be a good idea to mention what is said in the comment above, in the logging statement as well. (`it's assumed that the seeder is already running`)

Something like `Skipping startup of voucher seeder` or something could do.

julesdesmit: As a side note, I'm curious as to why you opted for assuming it is running, rather than failing if the path provided is not correct.

ggoranov: I think it could be helpful to support the case where voucher is already started manually. That would allow us to run E2E testing against testnet itself.

E.g another test file could define this situation:

- configure N local nodes (incl blindibid)
- connect to testnet seeder
- perform blockchain assertions at the point of testnet tip (not from genesis height)

julesdesmit: Ah, I see. In that case, I'll just ask you to add a little bit of info to the log entry. It might be a bit cryptic if another developer is running into trouble with the tests and can't tell from terminal output that it's only the seeder that isn't online/defined.

ggoranov: Noted.

ggoranov:
@julesdesmit  In the effort to merge test-harness with `automate-consensus`, I'm not sure how we could push a bid/stake/standard txs via jsonrpc as per latest changes. Note that neither CLI nor Transactor packages expose any RPCBus/EventBus interface.

A patch here could be to pass a ref of Cli instance to rpc pkg, but not ideal solution.

PR #78, opened 2019-09-06
Return Standard as a pointer, instead of a value

### LINKED ISSUE #77: Tx inputs are not added
Since the tx structures were unified, the `StandardTx` method was built to emulate the one from the old model, where it returns the `Standard` struct as a value. This will cause problems when creating transactions, as inputs are not saved on the correct `Standard` struct, and leaving the actual transaction bare.

julesdesmit:
Fixes #77

### DIFF
--- a/pkg/wallet/transactions/standardtx.go
+++ b/pkg/wallet/transactions/standardtx.go
@@ -266,6 +266,12 @@ func (s *Standard) prove(hasher func() ([]byte, error), encryptValues bool) erro
 	if err != nil {
 		return err
 	}
+	// Remove this has from the TxID field - the hash will be changed after the proof
+	// is done.
+	// TODO: CalculateHash should not set anything, as it's name implies only the
+	// calculation of the hash. This should be adjusted and the rest of the code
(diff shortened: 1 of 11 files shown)


pkg/wallet/transactions/standardtx.go
```diff
@@ -266,6 +266,12 @@ func (s *Standard) prove(hasher func() ([]byte, error), encryptValues bool) erro
 	if err != nil {
 		return err
 	}
+	// Remove this has from the TxID field - the hash will be changed after the proof
+	// is done.
+	// TODO: CalculateHash should not set anything, as it's name implies only the
+	// calculation of the hash. This should be adjusted and the rest of the code
```
ggoranov: If CalculateHash is refactored to not-set-txid idea, we'd need to ensure no perf regression occurs as a result. I used to observe perf impact in the past due to not-saved-txid and recalculating it here and there.

julesdesmit: Good point. I will open an issue for it and include your comment.

PR #84, opened 2019-09-13
Increase encoding efficiency 

### LINKED ISSUE #48: Refactor `encoding` package to increase efficiency
Currently there is room for improvement in passing by reference vs passing by value, especially in hotter parts of the code using this package (such as encoding and decoding events, transactions and big blocks).

julesdesmit:
Fixes #48. The benchmarks have not yet been ran on a proper target machine, but on my local machine the updates give us anywhere between 2 and 8 times speed increase, just by dropping interfaces and not passing pointers to read values into.

The dropping of interfaces removes the overhead of dynamic dispatch, which is probably our biggest efficiency gain as it is a costly operation. Aside from that, removing interfaces means that the buffers used for encoding/decoding can be allocated on the stack, saving us more time since this allocation is very cheap.


autholykos:
The change of encoding function signatures from `ReadXXX(r *bytes.Buffer, XXX) error` to `ReadXXX(r *bytes.Buffer) (XXX, error)` is a dubious one (in particular `ReadVarBytes`). Forcing allocations on API callers [has been pointed out as an anti-pattern in go](https://dave.cheney.net/2019/09/05/dont-force-allocations-on-the-callers-of-your-api), especially in performance optimization.

autholykos:
By the way, this is also the reason why the `Deserialize` function on the `UnMarshaller` objects have been suppressed

PR #91, opened 2019-09-18
TLS for RPC and GraphQl

vlopes11:
Implements #88


.travis.yml
```diff
@@ -3,6 +3,10 @@ dist: xenial
 
 language: go
 
+branches:
+  only:
+  - master
```
ggoranov: What's the motivation behind having Travis CI enabled only on master?
Wouldn't it cause an approved PR to break master on merging?

pkg/rpc/commands.go
```diff
@@ -38,12 +38,12 @@ var (
...
 // version will return the version of the client.
 var version = func(s *Server, params []string) (string, error) {
-	return protocol.NodeVer.String(), nil
+	return fmt.Sprintf("\"%v\"", protocol.NodeVer.String()), nil
 }
 
 // pong simply returns "pong" to let the caller know the server is up.
 var pong = func(s *Server, params []string) (string, error) {
-	return "pong", nil
+	return "\"pong\"", nil
```
ggoranov: In general, as discussed, JSONRPC should be kept for mutation/command requesting. Pong and version should be moved to graphql service. Maybe this can be handled by another PR.

vlopes11: Oh, cool, I'll do it now

.travis.yml
vlopes11: When the PRs target is master, travis will execute the full build/test, normally.

This ofc can be removed, but maybe we want to target the checking only for the master?

pkg/rpc/rpcserver.go
```diff
@@ -68,8 +70,21 @@ func (s *Server) Start() error {
 		s.handleRequest(w, *r, isAdmin)
 	})
 
-	// Set up listener
-	l, err := net.Listen("tcp", "localhost:"+cfg.Get().RPC.Port)
+	// Generate self-signed certificate
+	tlsCert, err := cryptoUtils.GenerateTLSCertificate()
```
ggoranov: Wouldn't it make sense to  try to read TLS cert and key file from config as first step? If TLS config is missing, then generate and use self-signed cert?

vlopes11: We have a future TODO to provide certs from the voucher:

https://github.com/dusk-network/dusk-voucher/issues/2

.travis.yml
ggoranov: Ah I see, my misunderstanding. If @Fulviuus  is fine with this, I'm ok too.

pkg/rpc/rpcserver.go
ggoranov: If the snippet below works for you we can handle this with one MR.

```go
certFile := cfg.Get().RPC.CertFile
keyFile := cfg.Get().RPC.KeyFile

keypair, err := tls.LoadX509KeyPair(certFile, keyFile)
if err != nil {
	return nil, err
}
```

pkg/rpc/commands.go
vlopes11: Ping, uptime and version moved to GraphQL

pkg/rpc/rpcserver.go
vlopes11: Done

pkg/gql/httpserver.go
```diff
@@ -90,8 +96,50 @@ func (s *Server) Start() error {
...
+		tlsCert, err = cryptoUtils.GenerateTLSCertificate()
+		if err != nil {
+			return err
+		}
+	}
+
+	// Define TLS configuration
+	tlsConfig := tls.Config{
+		Certificates:       []tls.Certificate{*tlsCert},
+		InsecureSkipVerify: true,
```
ggoranov: `InsecureSkipVerify` concerns only the client TLS connection, right? 
Do we need it enabled here?

pkg/gql/httpserver.go
```diff
@@ -90,8 +96,50 @@ func (s *Server) Start() error {
...
+
+		// Generate self-signed certificate
+		tlsCert, err = cryptoUtils.GenerateTLSCertificate()
+		if err != nil {
+			return err
+		}
+	}
+
+	// Define TLS configuration
+	tlsConfig := tls.Config{
```
ggoranov: Should we also specify the TLS min version, as by default it's TLS 1.0?

```
// MinVersion contains the minimum SSL/TLS version that is acceptable.
 // If zero, then TLS 1.0 is taken as the minimum.
 MinVersion uint16 // Go 1.2
```

vlopes11: You are correct. the minimum required version for the voucher is 1.3

vlopes11: Done

pkg/gql/httpserver.go
vlopes11: Also correct, and fixed. Late we will have safe certs

pkg/gql/httpserver.go
```diff
@@ -16,6 +18,7 @@ import (
 	logger "github.com/sirupsen/logrus"
 
 	cfg "github.com/dusk-network/dusk-blockchain/pkg/config"
+	cryptoUtils "github.com/dusk-network/dusk-blockchain/pkg/util/crypto"
```
julesdesmit: Generally, it is more idiomatic to import packages with all lowercase. So it should be imported as `cryptoutils`

vlopes11: Fixed

julesdesmit:
Closed due to inactivity and a growing amount of conflicts. Victor has said he will reimplement it later on, after finishing his current work.

PR #92, opened 2019-09-19
Issue 86

### LINKED ISSUE #86: `Transactor` should be decoupled from CLI and exposed over RPCBus
This would help ease E2E testing, as well as faucet functionality for the testnet.

Pending sub-tasks:

- [x] 1. Transactor to be started at the node loading procedure
- [x] 2. Transactor to unmarshal rpcBus params for all APIs
- [x] 3. RPCBus to support noTimeout calls
- [x] 4. CLI package to send rpcBus calls to Transactor (transferCMD, sendBidCMD, sendStakeBid, BalanceCMD, createFromSeedCMD, loadWallet, createWallet)
- [x] - Merge with master (solve conflicts)
- [x] 5. RPC package to send rpcBus calls to Transactor ( sendBidTx )
- [ ] 6. Update wallet on acceptedBlock topic
- [x] 7. Consensus maintainer  to send rpcBus calls to Transactor (SendBidTx, SendStakeTx)
- [x] 8. Initiator must send RoundUpdateTopic instead of InitializeTopic

### LINKED ISSUE #98: Delete from walletDB the associated keyImage on spending an output

### LINKED ISSUE #99: Initiator no longer starts consensus
With the merge of #93, the initiator was not properly updated to send a round update, instead of a initialization message. This will cause consensus to never start as nobody is currently listening to this message.

The initiator has no access to the consensus state, so it would probably be best if the chain listens for this message now, so that it can send the round update.

ggoranov:
Fixes #86, fixes #99, fixes #98


pkg/core/transactor/commands.go
```diff
@@ -0,0 +1,225 @@
...
+	err = t.w.Sign(tx)
+	if err != nil {
+		return nil, err
+	}
+
+	return tx, nil
+}
+
+func (t *Transactor) CreateStakeTx(amount, lockTime uint64) (transactions.Transaction, error) {
+	if err := t.syncWallet(); err != nil {
```
julesdesmit: If we are now syncing with each accepted block, can this call be removed from all tx creation methods, and the balance method?

pkg/core/transactor/transactor.go
```diff
@@ -2,191 +2,115 @@ package transactor
...
-		}
-		return nil
-	})
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return blk, state.TipHash, nil
+// Collector implements the wire.EventCollector interface
+type Collector struct {
```
julesdesmit: We already have an `AcceptedBlockCollector` in the consensus package I believe, so this is redundant.

ggoranov: If AcceptedBlockCollector can be used as a common util by different packages, it should be out of consensus pkg. 
I suggest we have another issue to make all pkgs that are collecting acceptedBlock use AcceptedBlockCollector from `utils` pkg.

pkg/core/transactor/commands.go
ggoranov: I decided to leave it for the situations where walletDB is moved to another node and a wallet call is sent before next acceptedBlock is accepted. To handle such corner case we could provide CLI command syncWallet. 
If you insist, we could ignore this and remove syncWallet for now.

pkg/core/transactor/transactor.go
julesdesmit: That sounds fine to me, but is there a reason you wrote this extra code? We should probably limit the amount of stuff to fix after this PR is merged, as we already have some skipped tests.

pkg/core/transactor/commands.go
julesdesmit: I'm not sure of a common scenario where we would move the walletDB out of the executable folder (in which case it can also just be copied). Thanks to the accepted block update we can always be sure we are synced to whatever our local chain tip is, so I believe this would be redundant.

ggoranov: Cool. Removing it. 
In this case we might need to document that after calling loadwallet, user will need to wait for the next acceptedBlock before sending any other wallet requests (bid,stake,transfer, balance)

julesdesmit: The wallet will still be synced with the last block it accepted on startup. But it's a good idea to maybe mention that wallet calls may fail as the node may be syncing when it's just started.

ggoranov: Note that this will break the E2E as it uses wallets with outputs in genesis block but the block at height 0 is never populated as acceptedBlock.

julesdesmit: Hmm, I hadn't considered it. Could we propagate the chain tip after setting up the chain?

PR #94, opened 2019-09-24
Pass event bus buffers by value

### LINKED ISSUE #61: Pass buffers in the Event Bus by value
Currently, to avoid sharing the same pointer, we copy the buffer being sent over the event bus for each handler in any given map of handlers. We can instead pass these buffers by value. This would remove the need for the `copyBuffer` function, and most likely make the dissemination of event buffers much more efficient.

julesdesmit:
Fixes #61


julesdesmit:
Closing, as #109 takes care of the issue in a more elegant way.

PR #100, opened 2019-09-26
Restore capability to launch consensus

### LINKED ISSUE #99: Initiator no longer starts consensus

julesdesmit:
Fixes #99

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -118,11 +118,10 @@ func (c *Chain) Listen() {
 func (c *Chain) LaunchConsensus() {
 	initChan := make(chan *bytes.Buffer, 1)
 	id := c.eventBus.Subscribe(msg.InitializationTopic, initChan)
-	roundBuf := <-initChan
-	round := binary.LittleEndian.Uint64(roundBuf.Bytes())
+	<-initChan
 	c.mu.RLock()
 	defer c.mu.RUnlock()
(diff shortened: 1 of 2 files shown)


pkg/core/chain/chain.go
```diff
@@ -113,6 +113,19 @@ func (c *Chain) Listen() {
...
+// LaunchConsensus will listen for an init message, and send a round update
+// once it is received.
+func (c *Chain) LaunchConsensus() {
+	initChan := make(chan *bytes.Buffer, 1)
+	id := c.eventBus.Subscribe(msg.InitializationTopic, initChan)
+	roundBuf := <-initChan
+	round := binary.LittleEndian.Uint64(roundBuf.Bytes())
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	c.sendRoundUpdate(round, c.prevBlock.Header.Seed, c.prevBlock.Header.Hash)
```
ggoranov: Instead of getting round from `msg.InitializationTopic`,  would it make sense to use `c.prevBlock.Header.Height + 1` ?

julesdesmit: Ah, good point. Would save us the need to actually read the buffer. Will adjust it

julesdesmit:
Closing, as this code is now included in #92

PR #109, opened 2019-10-02
Eventbus refactor

### LINKED ISSUE #61: Pass buffers in the Event Bus by value

autholykos:
This PR introduces substantial improvements to the architecture of the `EventBus` and lays the groundwork for further improvements on the consensus components. Incidentally fixes #61 and is a prerequisite for #93


pkg/util/nativeutils/eventbus/eventbus.go
```diff
@@ -1,280 +1,49 @@
...
 
-	"github.com/dusk-network/dusk-blockchain/pkg/util/container/ring"
 	lg "github.com/sirupsen/logrus"
 )
 
 var _ Broker = (*EventBus)(nil)
 
-var ringBufferLength = 2000
 var napTime = 1 * time.Millisecond
 var _signal struct{}
```
julesdesmit: These two variables could be removed, as they are no longer used.

pkg/util/nativeutils/eventbus/publisher.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (bus *EventBus) publish(topic string, event bytes.Buffer) {
+
+	// first serve the default topic listeners as they are most likely to need more time to (pre-)process topics
+	go bus.defaultListener.Notify(topic, event)
+
+	if listeners := bus.listeners.Load(topic); listeners != nil {
+		for _, listener := range listeners {
+			if err := listener.Notify(event); err != nil {
+				logEB.WithField("topic", topic).Warnln("listener failed to notidy buffer")
```
julesdesmit: notify*

julesdesmit:
LGTM, just seems that there's still some issue merging this in as the `master` branch has changed a bit since this refactor was started (a channel in the Chain has not been updated to work with the new API).

autholykos:
The merge problems should be fixed

PR #121, opened 2019-10-09
Deprecate topic listener

### LINKED ISSUE #120: Deprecate `TopicListener` and remove excessive channel usage from consensus components
This is preparatory work for #95 and it targets exclusively consensus components

autholykos:
fixes #120


julesdesmit:
LGTM

PR #145, opened 2019-11-19
Minor adjustments to codebase

julesdesmit:
- Logging level on failure to verify a block or certificate downgraded from error to warning, as it is not a severe issue. This should also prevent pollution of the monitoring error logs with errors that do not matter.
- Remove the pre-built blindbid binaries, as they don't belong in this repository. Pre-built blindbid binaries can be included in release tarballs instead.


autholykos:
LGTM

PR #157, opened 2019-11-25
Refuse transactions which contain locked inputs

### LINKED ISSUE #147: Mempool currently allows for spending locked inputs
When joining the current Devnet with a node, and spending the entire balance on consensus transactions which would be locked for 250,000 blocks, I was still able to spend my coins a short while afterwards.

julesdesmit:
Fixes #147 

Updates `dusk-wallet` from v0.1.2 to v1.0.1.

### DIFF
--- a/pkg/core/database/heavy/transactions.go
+++ b/pkg/core/database/heavy/transactions.go
@@ -163,7 +163,7 @@ func (t transaction) StoreBlock(b *block.Block) error {
 			// Only lock the first output, so that change outputs are
 			// not affected.
 			if i == 0 {
-				binary.LittleEndian.PutUint64(value, tx.UnlockHeight()+b.Header.Height)
+				binary.LittleEndian.PutUint64(value, tx.LockTime()+b.Header.Height)
 			}
 			t.put(append(OutputKeyPrefix, output.PubKey.P.Bytes()...), value)
 		}
--- a/pkg/core/database/heavy/transactions.go
+++ b/pkg/core/database/heavy/transactions.go
@@ -159,6 +159,7 @@ func (t transaction) StoreBlock(b *block.Block) error {
 		//
 		// To make FetchOutputKey functioning
 		for i, output := range tx.StandardTx().Outputs {
+			value := make([]byte, 8)
 			// Only lock the first output, so that change outputs are
 			// not affected.
 			if i == 0 {
--- a/pkg/core/database/heavy/transactions.go
+++ b/pkg/core/database/heavy/transactions.go
@@ -155,11 +155,13 @@ func (t transaction) StoreBlock(b *block.Block) error {
 		// Schema
 		//
 		// Key = OutputKeyPrefix + tx.output.PublicKey
-		// Value = tx.output.PublicKey
+		// Value = unlockheight
 		//
 		// To make FetchOutputKey functioning
 		for _, output := range tx.StandardTx().Outputs {
-			t.put(append(OutputKeyPrefix, output.PubKey.P.Bytes()...), output.PubKey.P.Bytes())
+			value := make([]byte, 0, 8)
+			binary.LittleEndian.PutUint64(value, tx.UnlockHeight())
+			t.put(append(OutputKeyPrefix, output.PubKey.P.Bytes()...), value)
 		}
 
 	}
@@ -247,6 +249,19 @@ func (t transaction) FetchOutputExists(destkey []byte) (bool, error) {
 	}
 
 	return exists, err
+}
+
+// FetchOutputUnlockHeight returns the unlockheight of an output
+func (t transaction) FetchOutputUnlockHeight(destkey []byte) (uint64, error) {
+	key := append(OutputKeyPrefix, destkey...)
+	unlockHeightBytes, err := t.snapshot.Get(key, nil)
+	if err != nil {
+		return 0, err
+	}
+
+	// output unlock height is the first 8 bytes
+	unlockHeight := binary.LittleEndian.Uint64(unlockHeightBytes[0:8])
+	return unlockHeight, err
 }
 
 // FetchDecoys iterates over the outputs and fetches `numDecoys` amount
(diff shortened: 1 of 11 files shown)


pkg/core/database/heavy/transactions.go
```diff
@@ -155,11 +155,13 @@ func (t transaction) StoreBlock(b *block.Block) error {
 		// Schema
 		//
 		// Key = OutputKeyPrefix + tx.output.PublicKey
-		// Value = tx.output.PublicKey
+		// Value = unlockheight
 		//
 		// To make FetchOutputKey functioning
 		for _, output := range tx.StandardTx().Outputs {
-			t.put(append(OutputKeyPrefix, output.PubKey.P.Bytes()...), output.PubKey.P.Bytes())
+			value := make([]byte, 8)
```
ggoranov: @julesdesmit Would you clarify the scenario from below?  What would happen if ...

0. Alice has 100 Dusk (in a single output) 
1. Alice sends a BidTx with amount of 10 Dusk, unlockHeight 55
2. As a result a tx with one input and two outputs is created
	- output_1 - 10 dusk (bid amount)
	- output_2 - 90 dusk (the change)
3. Both outputs will be stored into db key-value entries as
	output_1.PubKey, unlockHeight: 55
	output_2.PubKey, unlockHeight: 55

If the above is correct understanding, then Alice will not be capable of spending the `change` of 90 dusk until 55 height. So in that case even the bid amount is only 10 the entire bucket of 100 DUSK is locked.

Lemme know if I misunderstood it.

pkg/core/database/interfaces.go
```diff
@@ -94,6 +94,8 @@ type Transaction interface {
 
 	FetchOutputExists(destkey []byte) (bool, error)
 
+	FetchOutputUnlockHeight(destkey []byte) (uint64, error)
```
ggoranov: Could we add a bit of description/comments for each exported interface method?

pkg/core/database/interfaces.go
```diff
@@ -94,6 +94,8 @@ type Transaction interface {
...
+
 	SaveBidValues([]byte, []byte) error
```
ggoranov: not related to this PR.
Could we use Fetch/Store as prefixes to be aligned with the naming convention for this pkg?

pkg/core/database/lite/transactions.go
```diff
@@ -202,8 +202,12 @@ func (t transaction) FetchDecoys(numDecoys int) []ristretto.Point {
 func (t transaction) FetchOutputExists(destkey []byte) (bool, error) {
 	return false, nil
 }
-func (t *transaction) StoreCandidateBlock(b *block.Block) error {
 
+func (t transaction) FetchOutputUnlockHeight(destkey []byte) (uint64, error) {
```
ggoranov: Would it be possible to keep lite driver up-to-date so that it can be used for testing and benchmark purposes? Not urgent , can be done by another PR.

pkg/core/database/heavy/transactions.go
```diff
@@ -249,6 +251,19 @@ func (t transaction) FetchOutputExists(destkey []byte) (bool, error) {
 	return exists, err
 }
 
+// FetchOutputUnlockHeight returns the unlockheight of an output
```
ggoranov: In both dusk-blockchain (incl CLI help) and dusk-wallet, we use mostly `lockTime/lock`. Would it make sense to refactor this and use everywhere UnlockHeight to make it more readable, unless we are going to support both lock-by-height and lock-by-timestamp?

ggoranov:
Ideally, it could be good to have a E2E scenario in TestHarness for each major feature (like tx locking). If this is too much effort for this PR, let's just have a github issue.

julesdesmit:
Yes I agree, setting up the test was a bit of a hassle and it would be easier tested with the test harness. This would probably be a good test to add once we incorporate the test harness into the CI, so I will open up the issue for now.

pkg/core/database/heavy/transactions.go
julesdesmit: Good point. I think, with Phoenix, we will only support lock-by-height, so it would probably make sense to refactor.

pkg/core/database/heavy/transactions.go
julesdesmit: That's a really good catch. This is currently a bit tricky to work around, since the unlock height is specified on the transaction, and not on a per-output basis. Maybe a slight refactoring of the tx structure is in order.

julesdesmit: I think we could work around it by only locking the first output of any tx, as we currently do not support multi-output txs.

pkg/core/database/interfaces.go
julesdesmit: Good point, will add it.

pkg/core/database/interfaces.go
julesdesmit: Fair point

pkg/core/database/lite/transactions.go
julesdesmit: Also a good point. I will open an issue.

julesdesmit:
> Do you think it could be informative if we show Unlock Height in block explorer per each Bid/Stake transaction?

Sounds like a good idea to me.

pkg/core/database/heavy/transactions.go
julesdesmit: Your question actually got me to realize a terrible error I made: LockTime is not the same as UnlockHeight. I've adjusted the naming of these functions/fields and how they are writting into their databases for both `dusk-blockchain` and `dusk-wallet`. Once dusk-blockchain/dusk-wallet#19 is merged, I can finalize this fix. Hopefully, the naming should then no longer be confusing.

For clarity:
- `LockTime` is the amount of blocks to lock an output up for
- `UnlockHeight` is the absolute blockchain height at which the output becomes unlocked

PR #166, opened 2019-12-02
Implement correct certificate creation procedure, and the catch-up protocol

### LINKED ISSUE #24: Help nodes falling-behind to catch up
Due to some unexplained issue, some nodes during deployment can fall behind and never sync up with the rest of their peers. My hunch is that it is due to some nodes not accepting every block after finishing consensus, which ends up in a gap in their blockchain, and then when someone asks to sync with them, they only supply about half the blocks and leave the requesting node hanging. Need to do more testing to confirm this.

To quote @toghrulmaharramov 
> we have to work on formalizing a bootstrapping model for the nodes that have fallen behind for less than `n` blocks. Unfortunately, due to the complexity of `Reduction phase`, which muddles up the bootstrapping process, the catch-up process has to be made more complex than on rival protocols. I have a few ideas on how the problem could be elegantly curtailed

### LINKED ISSUE #118: Implement upgraded certificate creation procedure
Since we dropped the signature set loop, the decision of which certificate to use now rests, in a sense, on the block generators. Block generators are supposed to listen for agreement events, and once agreement is reached, pack these up into a certificate, which is included in the `Candidate` message for the next round. Whichever block candidate ends up winning that next round, that included certificate is used for the block which came before it. 

Outline of the certificate procedure:
Let `r` be a round in the consensus execution. When round `r` is finalized and a block is converged on, the certificate is created. This certificate then needs to be cached, and included on the next `candidate` message, sent by this same node. The certificate included with the candidate block that wins in round `r+1` will be attributed to the block finalized in round `r`. The full block is then accepted to the chain and propagated to peers.

julesdesmit:
Fixes #24 and fixes #118

### DIFF
--- a/pkg/core/chain/candidate/blockgenerator.go
+++ b/pkg/core/chain/candidate/blockgenerator.go
@@ -36,7 +36,7 @@ type Generator struct {
 	rpcBus    *rpcbus.RPCBus
 	signer    consensus.Signer
 
-	roundInfo    consensus.RoundUpdate
+	roundInfo    consensus.RoundState
 	scoreEventID uint32
 }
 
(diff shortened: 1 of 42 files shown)


julesdesmit:
Closing this PR, as I'm re-attempting the implementation

Issue #169, opened 2019-12-03
Mempool should prioritize txs based on StandardTx().Fee value


ggoranov:

ggoranov:
Using sort.Slice at each pool.Put call performs much slower than with disabled sorting.
Results below:

Disabled `sort.Slice` sorting
``` 
BenchmarkPut-8                 1        5344484857 ns/op
--- BENCH: BenchmarkPut-8
    hashmap_test.go:126: Pool transactions count: 30000
PASS
ok      github.com/dusk-network/dusk-blockchain/pkg/core/mempool        5.417s

```

Enabled  `sort.Slice` sorting

```
--- BENCH: BenchmarkPut-8
    hashmap_test.go:126: Pool transactions count: 30000
PASS
ok      github.com/dusk-network/dusk-blockchain/pkg/core/mempool        50.749s

```

BinarySearch  and data shift without sort.Slice

```
BenchmarkPut-8                 1        6305648889 ns/op
--- BENCH: BenchmarkPut-8
    hashmap_test.go:126: Pool transactions count: 30000
PASS
ok      github.com/dusk-network/dusk-blockchain/pkg/core/mempool        6.379s

```

autholykos:
Closed via #171

Issue #168, opened 2019-12-03
GetMempoolTx should be extended to support limit by size


ggoranov:
Currently, GetMempoolTxs returns up to 50 txs per request (hard-coded limit). 
Block Generator should be able to fetch a various number of transactions based on their total size.

autholykos:
Merged upstream via #170

Issue #167, opened 2019-12-03
Extend mempool logging


ggoranov:
Mempool logs should record:

- size of submitted tx
- time spent on processing tx

autholykos:
Merged upstream via #170

Issue #165, opened 2019-12-02
Create request-response routine for submiting mempool transactions


ggoranov:

autholykos:
Merged with upstream in #170

PR #170, opened 2019-12-03
Feature mempool refactor

ggoranov:
Fixes:
- https://github.com/dusk-network/dusk-blockchain/issues/165
- https://github.com/dusk-network/dusk-blockchain/issues/167
- https://github.com/dusk-network/dusk-blockchain/issues/168
- https://github.com/dusk-network/dusk-blockchain/issues/169

### DIFF
--- a/pkg/core/mempool/mempool.go
+++ b/pkg/core/mempool/mempool.go
@@ -44,8 +44,9 @@ var (
 // Mempool is a storage for the chain transactions that are valid according to the
 // current chain state and can be included in the next block.
 type Mempool struct {
-	getMempoolTxsChan <-chan rpcbus.Request
-	sendTxChan        <-chan rpcbus.Request
+	getMempoolTxsChan       <-chan rpcbus.Request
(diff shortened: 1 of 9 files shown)


pkg/core/mempool/mempool.go
```diff
@@ -124,53 +137,66 @@ func (m *Mempool) Run() {
...
-
 			}
 		}
 	}()
 }
 
-// onPendingTx ensures all transaction rules are satisfied before adding the tx
-// into the verified pool
+// onPendingTx handles a submitted tx from any source (rpcBus or eventBus)
 func (m *Mempool) onPendingTx(t TxDesc) ([]byte, error) {
```
autholykos: `onPendingTx` should not return values since errors are handled directly in the method body and the `txid` is never checked anywhere by the client of the function

ggoranov: `onPendingTx` return values are needed by `onSendMempoolTx` to enable rpcBus call to retrieve txid or error on submitting a tx to the pool.

PR #174, opened 2019-12-04
Substitute Ed25519 message signing and verification for Blake2b checksums

### LINKED ISSUE #172: Ed25519 signing should be replaced by hashing messages, and including checksums in network message headers
Going the extra mile to sign messages is not necessary if we add checksums to network message headers, so that we can ensure integrity. This will also get rid of some of the complexity in the code related to the consensus.

julesdesmit:
Fixes #172

### DIFF
--- a/pkg/p2p/peer/peer.go
+++ b/pkg/p2p/peer/peer.go
@@ -2,12 +2,14 @@ package peer
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"net"
 	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/sha3"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/core/database/heavy"
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/peer/dupemap"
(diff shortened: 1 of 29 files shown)


pkg/p2p/peer/peer.go
```diff
@@ -245,10 +247,33 @@ func (p *Reader) ReadLoop() {
...
+				"error":   err,
+			}).Warnln("error reading message")
+			return
+		}
+
+		p.router.Collect(bytes.NewBuffer(message))
 	}
 }
 
+func verifyChecksum(m []byte) ([]byte, error) {
```
autholykos: `verifyChecksum` should be moved into a `checksum` package and renamed into `checksum.Extract`. This is consistent with `topics.Extract` and the fact that it mutates the input (a function named `verifyXXX` should not mutate input parameters and should simply return a `bool` or an `error`)

pkg/p2p/peer/peer.go
```diff
@@ -261,13 +260,13 @@ func (p *Reader) ReadLoop() {
...
 
 func verifyChecksum(m []byte) ([]byte, error) {
-	// Last 4 bytes are the checksum
-	message := m[:len(m)-4]
-	checksum := m[len(m)-4:]
+	// First 4 bytes are the checksum
+	checksum := m[:processing.ChecksumLength]
+	message := m[processing.ChecksumLength:]
 
 	digest := sha3.Sum256(message)
```
autholykos: consider (i.e. benchmark) `Blake2` instead of `Sha3`

julesdesmit:
Both very good points, I will address them in the next few commits.

PR #177, opened 2019-12-05
Queue agreement events on a seperate queue which flushes on a round update

### LINKED ISSUE #176: Agreement messages are never pulled out of the queue
On the Agreement step in consensus, the queue is not emptied. If a node is one or more rounds behind, it means that these events will never be dispatched. Without a catch-up protocol, this can lead to the node falling behind indefinitely.

julesdesmit:
Fixes #176

### DIFF
--- a/pkg/core/consensus/round.go
+++ b/pkg/core/consensus/round.go
@@ -332,6 +332,19 @@ func (c *Coordinator) CollectEvent(m bytes.Buffer) error {
 		return nil
 	case header.After:
 		lg.WithField("topic", topic).Debugln("storing future event")
+
+		// If it is a future agreement event, we store it on the
+		// corresponding round for step 1. This would mean that the
+		// queued agreement events for that round will be dispatched
+		// at the start of the first selection.
+		if topic == topics.Agreement {
+			c.eventqueue.PutEvent(hdr.Round, 1, NewTopicEvent(topic, hdr, m))
+			c.lock.RUnlock()
+			return nil
+		}
+
+		// Otherwise, we just queue it according to the header round
+		// and step.
 		c.eventqueue.PutEvent(hdr.Round, hdr.Step, NewTopicEvent(topic, hdr, m))
 		c.lock.RUnlock()
 		return nil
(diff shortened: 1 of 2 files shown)


pkg/core/consensus/round.go
```diff
@@ -332,6 +332,19 @@ func (c *Coordinator) CollectEvent(m bytes.Buffer) error {
 		return nil
 	case header.After:
 		lg.WithField("topic", topic).Debugln("storing future event")
+
+		// If it is a future agreement event, we store it on the
+		// corresponding round for step 1. This would mean that the
+		// queued agreement events for that round will be dispatched
+		// at the start of the first selection.
+		if topic == topics.Agreement {
+			c.eventqueue.PutEvent(hdr.Round, 1, NewTopicEvent(topic, hdr, m))
```
autholykos: Rather than abusing the eventqueue, would it make sense to create a step-independent queue to be used alongside the step-dependent one? Forcing the step to `1` looks a bit hacky to me. Also, I am not really a fan of relying on a `selection` action to flush the queue. It is textbook _implicit side effect_ programming and is a sure way to regression

julesdesmit: You're right, that's a very good point. I will address this in the next commits.

pkg/core/consensus/round.go
```diff
@@ -334,11 +347,11 @@ func (c *Coordinator) CollectEvent(m bytes.Buffer) error {
 		lg.WithField("topic", topic).Debugln("storing future event")
 
 		// If it is a future agreement event, we store it on the
-		// corresponding round for step 1. This would mean that the
-		// queued agreement events for that round will be dispatched
-		// at the start of the first selection.
+		// `roundQueue`. This means that the event will be dispatched
+		// as soon as the Coordinator reaches the round in the event
+		// header.
 		if topic == topics.Agreement {
```
autholykos: I would be good to get rid of ad-hoc topic processing in the `Coordinator`, but I can live with this for now. Will create a separate issue

PR #205, opened 2019-12-19
Minor updates to the RPC server

### LINKED ISSUE #202: RPC Server should be automatically enabled
This makes it much easier for us to hook in front-end processes.

### LINKED ISSUE #203: RPC Server should be over TCP on localhost by default
Golang's `http` library does not support doing requests over unix domain sockets currently. It will be much easier to hook other processes in if this is defaulted to TCP.

### LINKED ISSUE #204: JSON Response `result` parameter should be a string, instead of a json.RawMessage
This makes it much easier to decode a response from the caller's perspective.

julesdesmit:
These changes will prep the RPC server to integrate with `dusk-wallet-cli` more easily.

Fixes #202, fixes #203, fixes #204

### DIFF
--- a/harness/engine/utils.go
+++ b/harness/engine/utils.go
@@ -69,7 +69,8 @@ func (n *Network) SendQuery(nodeIndex uint, query string, result interface{}) er
 	return nil
 }
 
-// SendCommand sends a jsonrpc request to the specified network node
+// SendCommand sends a jsonrpc request to the specified network node.
+// Returns a string with the response of the json-rpc server.
 func (n *Network) SendCommand(nodeIndex uint, method string, params []string) (string, error) {
 
 	if nodeIndex >= uint(len(n.Nodes)) {
--- a/harness/engine/utils.go
+++ b/harness/engine/utils.go
@@ -70,10 +70,10 @@ func (n *Network) SendQuery(nodeIndex uint, query string, result interface{}) er
 }
 
 // SendCommand sends a jsonrpc request to the specified network node
-func (n *Network) SendCommand(nodeIndex uint, method string, params []string) ([]byte, error) {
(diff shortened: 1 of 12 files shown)


harness/engine/utils.go
```diff
@@ -70,10 +70,10 @@ func (n *Network) SendQuery(nodeIndex uint, query string, result interface{}) er
 }
 
 // SendCommand sends a jsonrpc request to the specified network node
```
autholykos: add info about the semantics of the returned string. What should the client of `SendCommand` expect?

pkg/core/transactor/listener.go
```diff
@@ -119,6 +124,47 @@ func (t *Transactor) handleAddress(r rpcbus.Request) error {
...
+		s.WriteString(fmt.Sprintf("%.8f DUSK ", float64(record.Amount)/float64(wallet.DUSK)) + " / ")
+		// Unlock height
+		s.WriteString("Unlocks at " + strconv.FormatUint(record.UnlockHeight, 10) + " / ")
+		// Recipient
+		s.WriteString("Recipient: " + record.Recipient)
+
+		s.WriteString("\n")
+	}
+
+	r.RespChan <- rpcbus.Response{*bytes.NewBufferString(s.String()), nil}
```
autholykos: For now, this is OK. However, looking at the whole file, I suspect that tossing strings when returning values on the RPC bus is a code smell. I'd rather serialize a specific struct with a `String()` method. This way, if we'd ever need to migrate away from strings we would avoid the pain of backward incompatible behaviors

julesdesmit: Good point. I will put it in a separate issue.

PR #206, opened 2019-12-19
Add a bash script, and an executable, for launching both the blind bid and the node at the same time

### LINKED ISSUE #150: Create a small tool which starts both the blindbid and the node
This will make node usage a bit easier, and also ensure correct startup (for example, a user should not be able to start the node without the blindbid process if the `walletonly` flag is false, since this would lead to hangs).

julesdesmit:
Fixes #150

### DIFF
--- a/launch/start-node/start-node.go
+++ b/launch/start-node/start-node.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+// A very simple binary, tasked with starting the processes for both
+// the blind bid and the node.
+func main() {
+	blindBidCmd := exec.Command("./blindbid")
+	if err := blindBidCmd.Start(); err != nil {
+		fmt.Printf("could not start blind bid binary - %v\n", err)
+		os.Exit(1)
+	}
+
+	nodeCmd := exec.Command("./testnet")
+	if err := nodeCmd.Start(); err != nil {
+		fmt.Printf("could not start node binary - %v\n", err)
+	}
+}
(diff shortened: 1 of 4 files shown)


launch/start-node/start-node.go
```diff
@@ -0,0 +1,22 @@
...
+func main() {
+	blindBidCmd := exec.Command("./blindbid")
+	if err := blindBidCmd.Start(); err != nil {
+		fmt.Printf("could not start blind bid binary - %v\n", err)
+		os.Exit(1)
+	}
+
+	nodeCmd := exec.Command("./testnet")
+	if err := nodeCmd.Start(); err != nil {
+		fmt.Printf("could not start node binary - %v\n", err)
```
autholykos: In case of errors the program should return with an error code (like the above)

julesdesmit: Ah yeah, good catch. Totally forgot it there

launch/testnet/start-node.sh
```diff
@@ -0,0 +1,5 @@
+go build
+
+# Launch processes in the background, and ignore their output
+./blindbid &>/dev/null &
+./testnet &>/dev/null &
```
autholykos: in case of an unrecoverable error, the latter should be trapped, logged and the script should return an error code on exit

julesdesmit: The script runs the executables in the background, so that control is returned to the caller instantly (to then, for example, open the wallet). So the exit code can not be returned from the script itself.

In case of a panic though, maybe it makes sense to replace the `panic` statements and log these instead with `log.Panic`, so that they show up in the logfile.

PR #211, opened 2019-12-23
Make maintainer launch optional instead of mandatory

### LINKED ISSUE #156: Make stake/bid automation opt-in, rather than standard
Making the automation standard might confuse some users.

julesdesmit:
Fixes #156

### DIFF
--- a/pkg/core/consensus/initiator/initiator.go
+++ b/pkg/core/consensus/initiator/initiator.go
@@ -61,9 +61,11 @@ func storeBidValues(eventBroker eventbus.Broker, rpcBus *rpcbus.RPCBus, w *walle
 	_, db := heavy.CreateDBConnection()
 	for i := uint64(0); ; i++ {
 		hash, err := getBlockHashForHeight(db, i)
-		if err != nil {
+		if err == database.ErrBlockNotFound {
 			// We hit the end of the chain, so just exit here
 			return
+		} else if err != nil {
+			panic(err)
 		}
 
 		txs, err := getTxsForBlock(db, hash)
--- a/pkg/core/consensus/initiator/initiator.go
+++ b/pkg/core/consensus/initiator/initiator.go
@@ -2,12 +2,9 @@ package initiator
 
 import (
 	"bytes"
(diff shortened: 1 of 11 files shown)


pkg/core/consensus/initiator/initiator.go
```diff
@@ -54,35 +48,25 @@ func startProvisioner(eventBroker *eventbus.EventBus, rpcBus *rpcbus.RPCBus, w *
...
-			var err error
-			hash, err = t.FetchBlockHashByHeight(uint64(i))
-			return err
-		})
-
-		// We hit the end of the chain, so we should just exit here
+	for i := uint64(0); ; i++ {
+		hash, err := getBlockHashForHeight(db, i)
 		if err != nil {
+			// We hit the end of the chain, so just exit here
```
autholykos: We should strive to differentiate between errors associated with expected behaviors and unexpected errors. Please introduce an error type we can check (and handle differently)

pkg/core/consensus/maintainer/maintainer.go
```diff
@@ -141,20 +142,32 @@ func (m *StakeAutomaton) sendBid() error {
...
+	if err := rpcbus.MarshalConsensusTxRequest(buf, amount, lockTime); err != nil {
 		return err
 	}
 
 	_, err := m.rpcBus.Call(rpcbus.SendStakeTx, rpcbus.NewRequest(*buf), 0)
 	return err
 }
+
+func (m *StakeAutomaton) getTxSettings() (uint64, uint64) {
+	settings := config.Get().Consensus
```
autholykos: Assuming `settings` does not change at runtime, I would prefer to eliminate all references to `config` and inject it in the `StakeAutomaton` constructor by the client of this code

julesdesmit: I see. The idea was to retain some flexibility for the user, being able to adjust his values on the fly (as one's amount of DUSK can grow over time). If it is not desirable though, I can change it back to just take these values in the constructor.

PR #212, opened 2019-12-23
Add a JSON-RPC command to inquire whether or not the wallet is loaded

### LINKED ISSUE #210: Add an RPC command which tells the caller about the status of the wallet
This will tell front-end applications which menu to open.

julesdesmit:
Fixes #210

### DIFF
--- a/pkg/core/transactor/listener.go
+++ b/pkg/core/transactor/listener.go
@@ -419,12 +419,11 @@ func (t *Transactor) handleUnconfirmedBalance(r rpcbus.Request) error {
 }
 
 func (t *Transactor) handleIsWalletLoaded(r rpcbus.Request) error {
-	if t.w == nil {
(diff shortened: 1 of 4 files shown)


pkg/core/transactor/listener.go
```diff
@@ -416,6 +418,16 @@ func (t *Transactor) handleUnconfirmedBalance(r rpcbus.Request) error {
 	return nil
 }
 
+func (t *Transactor) handleIsWalletLoaded(r rpcbus.Request) error {
+	if t.w == nil {
+		r.RespChan <- rpcbus.Response{*bytes.NewBufferString("not loaded"), nil}
```
ggoranov: nitpick: It's more idiomatic if the producer (transactor) returns true/false as bool flag and consumers (rpc etc) translates this into a message ("not loaded"/"loaded")

julesdesmit: Gotcha, adding it.

PR #213, opened 2019-12-23
Add support for fetching synchronization progress as a percentage

### LINKED ISSUE #149: Add a synchronization progress bar to the user interface

julesdesmit:
Fixes #149

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -730,16 +730,16 @@ func (c *Chain) provideRoundResults(r rpcbus.Request) {
 }
 
 func (c *Chain) provideSyncProgress(r rpcbus.Request) {
-	c.mu.RLock()
-	prevBlockHeight := c.prevBlock.Header.Height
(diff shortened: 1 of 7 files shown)


pkg/core/chain/chain.go
```diff
@@ -713,6 +727,16 @@ func (c *Chain) provideRoundResults(r rpcbus.Request) {
 	r.RespChan <- rpcbus.Response{*buf, nil}
 }
 
+func (c *Chain) provideSyncProgress(r rpcbus.Request) {
+	c.mu.RLock()
+	prevBlockHeight := c.prevBlock.Header.Height
+	c.mu.RUnlock()
+
+	progressPercentage := (float64(c.highestSeen) / float64(prevBlockHeight)) * 100
```
ggoranov: We can have here a check for divide-by-zero panic. I think there are some edge cases this can be triggered.

julesdesmit: Good catch, I will add it.

PR #219, opened 2019-12-27
Disallow reading invalid round from request param

ggoranov:


autholykos:
LGTM. Not sure why codecov complains though

PR #227, opened 2020-01-02
Ensure coinbase on first intermediate block is deterministic

### LINKED ISSUE #226: First intermediate block should be completely deterministic
Currently, the coinbase transaction in the first intermediate block is still random between nodes. This can cause the nodes to have a different block from each other, which could be problematic when a node attempts to create a transaction with this block's coinbase as one of the decoys.

julesdesmit:
Fixes #226

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -3,6 +3,7 @@ package chain
 import (
 	"bytes"
 	"errors"
+	"math/big"
 	"time"
 
 	"encoding/binary"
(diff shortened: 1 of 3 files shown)


pkg/core/chain/chain.go
```diff
@@ -787,3 +780,19 @@ func mockFirstIntermediateBlock(prevBlockHeader *block.Header) (*block.Block, er
...
+	seed := make([]byte, 32)
+
+	keyPair := key.NewKeyPair(seed)
+	tx := transactions.NewCoinbase(make([]byte, 32), make([]byte, 32), 2)
+	var r ristretto.Scalar
+	r.SetZero()
+	tx.SetTxPubKey(r)
+
+	var reward ristretto.Scalar
+	reward.SetBigInt(big.NewInt(int64(config.GeneratorReward)))
```
autholykos: The reward should be a callback injected in this (and the other reward aware) functions, also to allow for dynamic reward adjusting.

autholykos: Can be also tackled in a separate issue/PR, as long as this is recorded

julesdesmit: Opened in #238

PR #232, opened 2020-01-04
Remove Preprocessor logic in preparation of EventBus modularization

### LINKED ISSUE #231: Remove preprocessors and reassign their logic to the proper component
The preprocessors should be eliminated from the architecture to be able to port the `EventBus` to a separate process without the complication for subscribers to deal with topic-wide preprocessors.

autholykos:
Fixes #231


pkg/util/nativeutils/republisher/republisher.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	l := eventbus.NewCallbackListener(r.Republish)
+	r.id = r.broker.Subscribe(r.tpc, l)
+	return r.id
+}
+
+// Republish intercepts an Agreement message and repropagates it immediately
+// after applying any eventual validation logic
+func (r *Republisher) Republish(b bytes.Buffer) error {
```
ggoranov: nitpick: Is the republisher intended only for use upon Agreement messages? It's also used for topics.Candidate and topics.Reduction.

julesdesmit: As far as I could tell, these are included. Check the reduction factory and candidate broker, and they should be there.

autholykos: no, indeed it is a copy/paste error

PR #242, opened 2020-01-13
Add functionality for getting a view of the mempool, with certain filters

### LINKED ISSUE #146: Add command which shows an overview of the mempool, with options to filter for specific items
The mempool overview should allow for filtering items based on:
- Address
- TxID
- Tx type

julesdesmit:
Fixes #146

### DIFF
--- a/pkg/core/mempool/hashmap.go
+++ b/pkg/core/mempool/hashmap.go
@@ -110,6 +110,13 @@ func (m *HashMap) Contains(txID []byte) bool {
 	return ok
 }
 
+// Get returns a tx for a given txID if it exists.
+func (m *HashMap) Get(txID []byte) transactions.Transaction {
+	var k txHash
+	copy(k[:], txID)
+	return m.data[k].tx
+}
+
 // Size of the txs
 func (m *HashMap) Size() uint32 {
 	return m.txsSize
--- a/pkg/core/mempool/hashmap.go
+++ b/pkg/core/mempool/hashmap.go
@@ -102,6 +102,19 @@ func (m HashMap) Clone() []transactions.Transaction {
 	return r
 }
 
+// FilterByType returns all transactions for a specific type that are
+// currently in the HashMap.
+func (m HashMap) FilterByType(filterType transactions.TxType) []transactions.Transaction {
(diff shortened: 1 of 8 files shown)


pkg/core/mempool/hashmap.go
```diff
@@ -110,6 +110,13 @@ func (m *HashMap) Contains(txID []byte) bool {
 	return ok
 }
 
+// Get returns a tx for a given txID if it exists.
+func (m *HashMap) Get(txID []byte) transactions.Transaction {
+	var k txHash
+	copy(k[:], txID)
+	return m.data[k].tx
```
ggoranov: Could we check here if key exists with ` txdesc, ok := m.data[k]`. Currently, it does not panic as map contains values but if someone changes it to map of references (like `data map[txHash]*TxDesc`) , this would be a hidden vulnerability making mempool panic.

julesdesmit: Good point

julesdesmit:
Have opened an issue.

pkg/core/mempool/mempool.go
```diff
@@ -408,6 +417,60 @@ func (m Mempool) onGetMempoolTxs(r rpcbus.Request) (bytes.Buffer, error) {
...
+		}
+
+		txs = append(txs, tx)
+	} else {
+		// In other cases, we will range through the hash map and pick out
+		// what we want depending on the filter.
+		txs = m.verified.Clone()
+
+		if len(r.Params.Bytes()) == 1 {
+			txs = filterTxsByType(txs, transactions.TxType(r.Params.Bytes()[0]))
```
ggoranov: nitpick: `Clone()`  iterates through all mempool elements. We can have another API `CloneWithFilter(..)` to avoid `filterTxsByType` array shifting:

Just a suggestion:
```go


func (m HashMap) CloneWithFilter(filterType transactions.TxType) []transactions.Transaction {

	r := make([]transactions.Transaction, 0)
	for _, t := range m.data {
		if t.tx.Type() == filterType {
			r = append(r, t.tx)
		}
	}

	return r
}
```

julesdesmit: Added it

PR #260, opened 2020-01-21
Finish implementation of light-node mode

### LINKED ISSUE #253: Add service flags, and distinguish between light and full node peers
This allows for light nodes to join the network, and it will help us to only forward/request messages to the correct nodes, as light nodes should not bother with consensus messaging and requests, and txs.

### LINKED ISSUE #259: Finish implementation of light-node mode
Due to the nature of our blockchain, a 'light client' will still need to fully store and verify blocks. Thus, we should expand the impact of the `walletonly` flag in the configuration, in order to disable those components which are not truly needed. As far as I am aware, these components are:
- Consensus
- Mempool

In addition, the used service flag should change depending on this flag, to avoid using up an unnecessary amount of bandwidth, and ignore requests which it can't respond to, when a node is only participating in light mode.

julesdesmit:
Fixes #259, fixes #253


docs/lightnode.md
```diff
@@ -0,0 +1,41 @@
+## Light-node mode
+
+The node offers an option for the user to start up the node in what's called 'light-node' mode. It is intended for those wishing to only use the wallet functionality, and disables certain components, and allows the node to ignore most of the network messages, drastically decreasing bandwidth consumption. Below follows a more in-depth explanation of the effects 'light-node' mode has on the process.
```
ggoranov: Is light-mode intended only for wallet nodes?  Could dusk-explorer node benefit from it too?

launch/testnet/server.go
```diff
@@ -44,8 +44,11 @@ func Setup() *Server {
 	// creating the rpcbus
 	rpcBus := rpcbus.New()
 
-	m := mempool.NewMempool(eventBus, rpcBus, nil)
-	m.Run()
+	// Light nodes don't need to keep a mempool
+	if !cfg.Get().General.WalletOnly {
```
ggoranov: Is `WalletOnly config` kept for backward compatibility? Wouldn't `General.LightNode` name make code more informative?

docs/lightnode.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Future work
+
+In a brief discussion that was had about the implementation, Matteo suggested to take a different, more extensible approach in the future, as having what's known as a 'branching codebase' is considered to be a code smell. The idea is simply, instead of including the service flag in the handshake message, including a bitmask of message topics which can be sent and received by the node in question. Going off of this bitmask, a receiving node can then accept messages which match it, or drop messages which don't. Similarly, a sending node can decide whether or not to send a certain message, by checking it against this bitmask. This would allow for greater extensibility of perhaps different node types in the future (the block explorer comes to mind), and rids us of the single config flag, which branches the codebase.
```
ggoranov: Indeed! If we need to introduce another type of `light-node` (e.g block explorer might need to display mempool status) the code branching might get more complicated.

Might be an idea for future to enable a node user to assemble a node from different components just by specifying a config file, instead of branching on a single flag.

pkg/core/chain/chain.go
```diff
@@ -197,11 +197,10 @@ func (c *Chain) Close() error {
...
 func (c *Chain) onAcceptBlock(m bytes.Buffer) error {
-	// Ignore blocks from peers if we are only one behind - we are most
-	// likely just about to finalize consensus.
-	// TODO: we should probably just accept it if consensus was not
-	// started yet
-	if !c.counter.IsSyncing() {
+	// Ignore blocks from peers if we are only one behind, and in
+	// full-node mode - we are most likely just about to finalize
+	// consensus.
+	if !cfg.Get().General.WalletOnly && !c.counter.IsSyncing() {
```
ggoranov: Would `cfg.Get().General.WalletOnly` here make an inconsistency between lightNode and fullNode?  To what (mis)understand full node would be one block behind the light node?

docs/lightnode.md
julesdesmit: Dusk-explorer might be able to benefit from it, though it limits access to certain information (no mempool, no consensus messaging).

launch/testnet/server.go
julesdesmit: True. Will adjust

pkg/core/chain/chain.go
julesdesmit: You need to take into account the way the consensus works. As I've outlined before, the reason a full node does not accept outside blocks until it is two behind has to do with the introduction of the `intermediate block` and the `sleeper` and `dozer` mode. A full node will not need to accept blocks from other peers until it is two behind, since it will accept these blocks internally as received and validated through the consensus. When it does fall behind this far, it should immediately sync up as far as possible and then request additional info. We do not want this procedure to happen until it is absolutely necessary, hence why we wait to be behind two blocks, as slight delays between message propagation can already cause it to happen when considering the case where a full node accepts blocks from other peers that are only 1 ahead.

A full node running consensus is in actuality (assuming it is at the tip of the execution trace) one block ahead of it's light node peers, as it has the intermediate block.

pkg/p2p/peer/peer.go
```diff
@@ -40,9 +40,16 @@ type Connection struct {
...
 	*Connection
+	protocol.ServiceFlag
 }
 
 func (g *GossipConnector) Write(b []byte) (int, error) {
+	// Preliminary check, to ensure we don't send messages to
+	// nodes who aren't interested in them.
+	if !g.canSend(b[0]) {
+		return 0, nil
+	}
```
autholykos: I would rather prefer the Gossip connector to be unaware of _whitelisting logic_ and, instead, instantiate different `GossipConnector`s depending on the peer attached. Also, checking the first byte is a hacky way to differentiate messages. If, in fact, we implement continuous streaming, the packets could be chopped off in several chunks and the first byte would not be the packet topic/category anymore

pkg/p2p/peer/peer.go
```diff
@@ -159,7 +159,7 @@ func (w *Writer) Serve(writeQueueChan <-chan *bytes.Buffer, exitChan chan struct
 
 	// Any gossip topics are written into interrupt-driven ringBuffer
 	// Single-consumer pushes messages to the socket
-	g := &GossipConnector{w.gossip, w.Connection}
+	g := &GossipConnector{w.gossip, w.Connection, serviceFlag}
```
autholykos: I would use the `serviceFlag` to instantiate different `peer.Writer`s instead of modifying the `GossipConnector` logic depending on the flag

pkg/p2p/peer/peer.go
julesdesmit: Yep, very good point. My approach is definitely a bit hacky, but I didn't consider that option. I'll make the change

julesdesmit: As for reading the first byte, I'm not sure how else to approach it at the moment. One thing that came to mind is to do such a check in the `Gossip`, as I imagine it would tokenize the stream of bytes before sending it out to the peer, though I'm not sure if it's a good idea to give the `Gossip` the responsibility of deciding whether the payload is valid or not.

autholykos:
This branch has aged quite badly, and the amount of conflicts to be solved makes it more convenient to reapply the modifications directly on a more up-to-date branch

PR #262, opened 2020-01-21
Centralize message

### LINKED ISSUE #82: `CalculateHash` should only return the hash, and not also set it on the tx struct
This caused some confusion before, while trying to get integration tests working. The hash was set before the rangeproof was created, and as a result, the eventual hash differed from the one on the `TxID` field. It is currently just directly unset after calculating it while proving the rangeproof, as a quick patch, but as the name implies, maybe it would make more sense to only return the hash, and let the caller choose whether or not to set it.

### LINKED ISSUE #237: Replace the bytes.Buffer based message passing for the empty interface{}, coupled with type assertions
As a benchmark shows, when passing messages using the `bytes.Buffer` and encoding/decoding the data, we lose quite a bit of time.

```
goos: linux
goarch: amd64
pkg: github.com/dusk-network/dusk-blockchain/pkg/core/marshalling
BenchmarkEncodeDecodeBlock-8           3         347057829 ns/op
BenchmarkTypeCastBlock-8        84495676                14.0 ns/op
PASS
```

Thus, substituting the internal encoding/decoding of messages for a simple `interface{}` with a type assertion might increase our performance drastically.

### LINKED ISSUE #270: Avoid signature copying on verification
After the fix to dusk-network/bn256#6 there is no longer any need to copy signature slices as the `bn256.Decompress` (intrinsic within `msg.VerifyBLSSignature`) does not mutate the input

autholykos:
Fixes #237, fixes #82, fixes #270


pkg/util/nativeutils/eventbus/mock.go
```diff
@@ -61,6 +61,60 @@ func CreateFrameStreamer(topic topics.Topic) (*EventBus, io.WriteCloser) {
...
+		Reader: bufio.NewReader(r),
+		Writer: bufio.NewWriter(w),
+	}
+}
+
+// Write receives the packets from the ringbuffer and writes it on the internal
+// pipe immediatelyh
+func (sss *StupidStreamer) Write(p []byte) (n int, err error) {
+	lg := make([]byte, 4)
+	binary.LittleEndian.PutUint16(lg, uint16(len(p)))
```
julesdesmit: Is there a reason you're using a `uint16` here over a `uint32`? The `uint32` would utilize all bytes, and allow for testing with larger packets if needed in the future.

pkg/util/nativeutils/eventbus/listener.go
```diff
@@ -125,21 +132,17 @@ func newMultiListener() *multiListener {
...
-	if m.Has([]byte{byte(topic)}) {
-		// creating a new Buffer carrying also the topic
-		tpcMsg := topic.ToBuffer()
-		if _, err := tpcMsg.ReadFrom(&r); err != nil {
-			log.WithField("topic", topic.String()).WithError(err).Warnln("error in writing topic to a multi-dispatched packet")
-			return
-		}
+func (m *multiListener) Forward(topic topics.Topic, msg message.Message) {
+	if !m.Has([]byte{byte(topic)}) {
+		// TODO: maybe log this
```
julesdesmit: Not sure if that's necessary, since we would then effectively add a log entry for every single message not intended for the `multiListener`, where it's already sort of implied that the `multiListener` doesn't care about it.

pkg/rpc/commands.go
```diff
@@ -98,7 +101,7 @@ var sendBidTx = func(s *Server, params []string) (string, error) {
 		return "", fmt.Errorf("missing parameters: amount/locktime")
 	}
 
-	amount, err := strconv.Atoi(params[0])
+	amount, err := strconv.Atoi(params[1])
```
julesdesmit: This doesn't seem right

pkg/p2p/wire/message/reduction.go
```diff
@@ -0,0 +1,208 @@
...
+
+	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/header"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/encoding"
+	"github.com/dusk-network/dusk-blockchain/pkg/util"
+	"github.com/dusk-network/dusk-crypto/bls"
+	"github.com/dusk-network/dusk-wallet/v2/key"
+)
+
+type (
+	// the consensus.
```
julesdesmit: This comment looks a bit unfinished. :)

pkg/core/consensus/reduction/firststep/reducer.go
```diff
@@ -127,47 +124,72 @@ func (r *Reducer) sendReduction(step uint8, hash []byte) {
...
+// the Consensus components
+type StepVotesMsgFactory struct {
+	sv   message.StepVotes
+	hash []byte
+}
+
+// Create a new StepVotesMsgFactory
+func (s StepVotesMsgFactory) Create(sender []byte, round uint64, step uint8) consensus.InternalPacket {
+	// handling the case if the StepVotes is empty
+	// TODO: interface - check this. The step sounds like it is intrinsic
```
julesdesmit: The step can essentially be left out of the StepVotes, if I'm not mistaken. With the consensus refactoring, steps progress in sequence between all of the components, and thus the reduction steps can be inferred. This is probably more appropraite in a separate issue though, and should be checked.

pkg/util/nativeutils/eventbus/mock.go
autholykos: Sure thing. Does not change much and would avoid subtle hard-to-detect bugs in case we would test with bulky packages. Will change

pkg/util/nativeutils/eventbus/listener.go
autholykos: Right. I was under the wrong impression that the topic check was operated by the `Publisher`, but this is not the case. Indeed the TODO can be removed

pkg/rpc/commands.go
autholykos: yeah, looks like a vim creep. Will change

pkg/core/consensus/round.go
```diff
@@ -257,12 +254,10 @@ func (c *Coordinator) CollectRoundUpdate(m bytes.Buffer) error {
...
 
-	// TODO: the Coordinator should not send events. someone else should kickstart the
+	// TODO: the Coordinator should not send events. Someone else should kickstart the
 	// consensus loop
-	c.store.Dispatch(TopicEvent{
-		Topic: topics.Generation,
-		Event: Event{},
-	})
+	// TODO: interface - check this
+	c.store.Dispatch(message.New(topics.Generation, nil))
```
julesdesmit: The nil payload causes a panic at line 123, when a typecast is attempted to `consensus.InternalPacket`.

pkg/core/consensus/selection/selector.go
```diff
@@ -138,30 +150,28 @@ func (s *Selector) IncreaseTimeOut() {
...
+	var bestEvent consensus.InternalPacket
 	s.eventPlayer.Pause(s.scoreID)
-	buf := new(bytes.Buffer)
 	s.lock.RLock()
-	bestEvent := s.bestEvent
+	bestEvent = s.bestEvent
 	s.lock.RUnlock()
+
 	// If we had no best event, we should send an empty hash
 	if bestEvent == nil {
```
julesdesmit: As we set the `bestEvent` to an `EmptyScore` when starting selection, this condition will never be true, and instead the selector will send a `bestScore` with everything zero'd out, which causes consensus to stall.

pkg/p2p/wire/message/reduction.go
autholykos: I am not sure what you have against postmodern expressionism in comments. Anyway, I will cha.

pkg/core/consensus/reduction/firststep/reducer.go
autholykos: this indeed needs to be addressed through a low-priority separate issue. Will open one

pkg/core/consensus/round.go
autholykos: Ops. Will change

pkg/core/consensus/selection/selector.go
autholykos: I think this was an oversight from my side. I introduced the `Score.IsEmpty` method appositely as my intention is to get rid of as many `nil` check as we can. Please double check if changing to the following would implement the intended behavior:
```golang
if bestEvent.IsEmpty() {
    bestEvent = s.signer.Compose(emptyScoreFactory{})
}
```

julesdesmit: Yep, that should be fine, just tested it on the harness. (Although did need to typecast `bestEvent`)

pkg/core/consensus/selection/selector.go
```diff
@@ -138,30 +150,28 @@ func (s *Selector) IncreaseTimeOut() {
...
-		s.signer.SendInternally(topics.BestScore, emptyScore[:], buf, s.ID())
-	} else {
-		s.signer.SendInternally(topics.BestScore, bestEvent.VoteHash, buf, s.ID())
+		bestEvent = s.signer.Compose(emptyScoreFactory{})
 	}
 
+	msg := message.New(topics.Score, bestEvent)
```
julesdesmit: This message topic should be `BestScore`, as it would not arrive at the first step reducer otherwise.

autholykos:
The tests fail at the moment since they depend on the upgrade of bn256 within dusk-crypto

pkg/core/consensus/selection/selector.go
autholykos: the message is sent under the `topics.BestScore` (`s.signer.SendInternally(topics.BestScore, msg, s.ID())`). However, the message category is that of a `Score` (pretty much like sending on `Gossip` any message would cause the category and topic to differ). Is this causing problems?

julesdesmit: Yup, as the `Coordinator` routes messages based on their `Message.Category`, and not the topic given to `SendInternally`.

autholykos: Gotcha

autholykos: I will correct it, but this seems like an architectural flaw. It forces messages to know where they will end up to and breaks message immutability. The example of `BestScore` and `Score` is quite emblematic in that respect. It would perhaps be better for the `multilistener` to reroute messages internally on the basis of their topic, rather than their category.

julesdesmit: That could work as well. Up to you

autholykos: I opened an issue about it as it could lead to more refactoring, which I wanna avoid within the context of this PR

pkg/core/consensus/selection/selector.go
```diff
@@ -138,30 +150,28 @@ func (s *Selector) IncreaseTimeOut() {
...
+	bestEvent = s.bestEvent
 	s.lock.RUnlock()
+
 	// If we had no best event, we should send an empty hash
-	if bestEvent == nil {
-		s.signer.SendInternally(topics.BestScore, emptyScore[:], buf, s.ID())
-	} else {
-		s.signer.SendInternally(topics.BestScore, bestEvent.VoteHash, buf, s.ID())
+	if bestEvent.IsEmpty() {
+		bestEvent = s.signer.Compose(emptyScoreFactory{}).(message.Score)
```
julesdesmit: It appears that the `emptyScoreFactory` composes a `ScoreProposal`, which causes another panic here.

pkg/core/consensus/generation/generator.go
```diff
@@ -47,6 +52,8 @@ func (g *Generator) ID() uint32 {
 }
 
 // Collect `Restart` events and triggers a Generation event
-func (g *Generator) Collect(ev consensus.Event) error {
-	return g.signer.SendInternally(topics.Generation, emptyHash[:], emptyPayload, g.ID())
+func (g *Generator) Collect(_ consensus.InternalPacket) error {
+	packet := g.signer.Compose(restartFactory)
+	msg := message.New(topics.Restart, packet)
+	return g.signer.SendInternally(topics.Generation, msg, g.ID())
```
julesdesmit: Due to the topic being `Generation`, but the message category being `Restart`, this causes a loop. The message will end back up in the same place, causing infinite dissemination of a `Restart` message. Message category should be set to `Generation` for as long as #272 is not fixed.

pkg/p2p/peer/router.go
```diff
@@ -54,47 +55,61 @@ func (m *messageRouter) CanRoute(topic topics.Topic) bool {
...
 	case topics.Inv:
-		err = m.dataRequestor.RequestMissingItems(b)
+		topics.Extract(&b)
+		err = m.dataRequestor.RequestMissingItems(&b)
 	case topics.Block:
-		err = m.synchronizer.Synchronize(b, m.peerInfo)
+		topics.Extract(&b)
+		err = m.synchronizer.Synchronize(&b, m.peerInfo)
 	case topics.Ping:
+		topics.Extract(&b)
```
julesdesmit: Could we instead, extract the topic before going into the `switch` statement? To avoid some duplicate code

pkg/p2p/peer/router.go
```diff
@@ -54,47 +55,61 @@ func (m *messageRouter) CanRoute(topic topics.Topic) bool {
...
 
-func (m *messageRouter) route(topic topics.Topic, b *bytes.Buffer) {
+func (m *messageRouter) route(b bytes.Buffer, msg message.Message) error {
 	var err error
-	switch topic {
+	category := msg.Category()
+	switch category {
 	case topics.GetBlocks:
-		err = m.blockHashBroker.AdvertiseMissingBlocks(b)
+		topics.Extract(&b)
```
julesdesmit: Do we need to extract the topic here as well? Since `message.Unmarshal` already extracts this topic.

pkg/p2p/wire/message/reduction.go
```diff
@@ -0,0 +1,209 @@
...
+		return err
+	}
+
+	m.SetPayload(bev)
+	return nil
+}
+
+// Unmarshal unmarshals the buffer into a Reduction event.
+func UnmarshalReduction(r *bytes.Buffer, bev *Reduction) error {
+	if err := header.Unmarshal(r, &bev.hdr); err != nil {
```
julesdesmit: When called from `UnmarshalReductionMessage`, the header is unmarshalled twice, leading to unnecessary EOF errors.

pkg/p2p/wire/message/message.go
```diff
@@ -0,0 +1,237 @@
...
+			sb.WriteString("<payload is non-empty but not a Stringer>")
+		}
+	}
+	sb.WriteString("\n]\n")
+	sb.WriteString("\n")
+	return sb.String()
+
+}
+
+func (m simple) Id() []byte {
```
julesdesmit: This method possibly mutates `m`, so it should have a pointer receiver. 
```
func (m *simple) Id() []byte {
```

pkg/p2p/wire/message/message.go
```diff
@@ -0,0 +1,237 @@
...
+		buf, err := Marshal(m)
+		if err != nil {
+			panic(err)
+		}
+		m.marshalled = &buf
+	}
+	return m.marshalled.Bytes()
+}
+
+func (m simple) setPayload(i interface{}) {
```
julesdesmit: Same here

pkg/p2p/wire/message/message.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (m simple) Category() topics.Topic {
+	return m.category
+}
+
+func (m simple) Payload() interface{} {
+	return m.payload
+}
+
+func (m simple) SetPayload(payload interface{}) {
```
julesdesmit: And here

pkg/core/consensus/reduction/firststep/reducer.go
```diff
@@ -89,24 +88,22 @@ func (r *Reducer) Finalize() {
 }
 
 // Collect forwards Reduction to the aggregator
-func (r *Reducer) Collect(e consensus.Event) error {
-	ev := reduction.New()
-	if err := reduction.Unmarshal(&e.Payload, ev); err != nil {
-		return err
-	}
+func (r *Reducer) Collect(e consensus.InternalPacket) error {
+	red := e.(message.Reduction)
```
julesdesmit: `Reduction` messages are unmarshalled and stored in a pointer when received from the network, so this typecast causes a panic. This might be the case for lots of messages out there, so it's probably a good idea to compare the unmarshalling logic with the associated typecast later down the line.

pkg/p2p/peer/peer.go
```diff
@@ -264,7 +266,9 @@ func (p *Reader) ReadLoop() {
 			return
 		}
 
-		p.router.Collect(bytes.NewBuffer(message))
+		// TODO: error here should be checked in order to decrease reputation
+		// or blacklist spammers
+		_ = p.router.Collect(message)
```
julesdesmit: Would be good to at least log it already, as I had some issues in `router.Collect` pass me by due to dropping this error

pkg/p2p/wire/message/message.go
```diff
@@ -0,0 +1,237 @@
...
+		*b = payload.(bytes.Buffer)
+
+	default:
+		if err := marshalMessage(s.Category(), payload, b); err != nil {
+			return err
+		}
+	}
+
+	topics.Prepend(b, s.Category())
+	return nil
```
julesdesmit: nitpick: `return topics.Prepend(b, s.Category())`

don't wanna drop that error value

pkg/p2p/wire/message/message.go
```diff
@@ -0,0 +1,237 @@
...
+	var buf bytes.Buffer
+	// if it is a simple message, first we check if this message carries a
+	// cache of its marshalled form first
+	m, ok := s.(simple)
+
+	if !ok { // it is not a cacheable message. We simply marshal without caring for any optimization
+		if err := marshal(s, &buf); err != nil {
+			return bytes.Buffer{}, err
+		}
+		return buf, nil
```
julesdesmit: I believe that cached buffers no longer include the topic byte (observed this with `Reduction` republishing). It should be added here, though needs to be double checked if this is the case for all messages.

julesdesmit:
This branch is currently being ran on the devnet. If no errors occur in the next 24 hours, I think it is fine to merge.

PR #265, opened 2020-01-22
Fix race condition between validating a candidate hash, and accepting a candidate block

### LINKED ISSUE #263: Race condition between validating a candidate hash and accepting a candidate message
As observed on the testnet, often times the candidate message will arrive to the candidate broker before the block hash is validated by the selector. This leads to unnecessary delays in reaching consensus.

### LINKED ISSUE #264: Candidate broker `validHashes` map should be thread-safe
Currently, the `validHashes` map is not protected for concurrent use, though it is used across multiple goroutines, as hashes are validated through a callback on the event bus,

### LINKED ISSUE #267: Mempool should be cleared both on the accepted block update, and the intermediate block update
On the testnet, some block generators were observed attempting to include the same transactions twice. This happened whenever the block generator went into sleeper mode and had to catch up a few blocks. Since the mempool is only cleared on the intermediate block, these transactions which should have been cleaned up, remain in their mempool, and inevitably find their way into a block.

julesdesmit:
Fixes #263, fixes #264, fixes #267

### DIFF
--- a/pkg/core/candidate/broker.go
+++ b/pkg/core/candidate/broker.go
@@ -3,6 +3,7 @@ package candidate
 import (
 	"bytes"
 	"errors"
+	"sync"
 	"time"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
@@ -27,6 +28,7 @@ type Broker struct {
 	republisher *republisher.Republisher
 	*store
 	// List of block hashes for which a valid Score message was seen.
+	lock        sync.RWMutex
 	validHashes map[string]struct{}
 
 	acceptedBlockChan <-chan block.Block
@@ -61,7 +63,10 @@ func (b *Broker) Listen() {
 	for {
 		select {
 		case cm := <-b.candidateChan:
-			if _, ok := b.validHashes[string(cm.Block.Header.Hash)]; ok {
+			b.lock.RLock()
+			_, ok := b.validHashes[string(cm.Block.Header.Hash)]
+			b.lock.RUnlock()
+			if ok {
 				b.storeCandidateMessage(cm)
 			}
 		case r := <-b.getCandidateChan:
@@ -79,7 +84,9 @@ func (b *Broker) AddValidHash(m bytes.Buffer) error {
 		return err
 	}
 
+	b.lock.Lock()
 	b.validHashes[string(hash)] = struct{}{}
+	b.lock.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/util/nativeutils/republisher/republisher.go
```diff
@@ -5,9 +5,10 @@ import (
 
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/topics"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/eventbus"
+	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/prerror"
 )
 
-type Validator func(bytes.Buffer) error
+type Validator func(bytes.Buffer) *prerror.PrError
```
autholykos: I somewhat dislike setting priorities during `error` creation. I get the impression that it is a practice that pollutes the interface by forcing implementers to handle information they should have no say about. Whether an error is high priority or not, is none of the Validator business, and it should be a responsibility left to the components that handle the validation errors. I would suggest to include some common validation errors (something like "DuplicatePayloadError", "EncodingError" and "InvalidError") and type-check/handle them at the caller level

pkg/core/candidate/broker.go
```diff
@@ -27,13 +26,14 @@ type Broker struct {
 	publisher   eventbus.Publisher
 	republisher *republisher.Republisher
 	*store
-	// List of block hashes for which a valid Score message was seen.
 	lock        sync.RWMutex
+	queue       map[string]*Candidate
```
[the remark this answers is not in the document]
julesdesmit: Good point, will adjust it.

pkg/util/nativeutils/republisher/republisher.go
julesdesmit: I see what you mean. I'll adjust it.

pkg/util/nativeutils/republisher/republisher.go
```diff
@@ -52,15 +60,9 @@ func (r *Republisher) Activate() uint32 {
 func (r *Republisher) Republish(b bytes.Buffer) error {
 	for _, v := range r.validators {
 		err := v(b)
-		if err != nil && err.Priority == prerror.High {
+		if err != nil && (err == InvalidError || err == EncodingError) {
```
autholykos: nitpicking, but I'd rather use a `switch` with a `default` rather than an `if`

PR #275, opened 2020-02-05
Unmarshal incoming block before propagating it over the event bus

### LINKED ISSUE #274: Chain synchronizer should send an unmarshalled block over the event bus
The current configuration can cause panics, as the `Chain` expects a `block.Block`, when a `bytes.Buffer` is sent.

julesdesmit:
Fixes #274

### DIFF
--- a/pkg/p2p/peer/processing/chainsync/sync.go
+++ b/pkg/p2p/peer/processing/chainsync/sync.go
@@ -87,10 +87,15 @@ func (s *ChainSynchronizer) Synchronize(blkBuf *bytes.Buffer, peerInfo string) e
 		return nil
 	}
 
+	// Does the block come directly after our most recent one?
 	if diff == 1 {
-		// Write bufio.Reader into a bytes.Buffer so we can send it over the event bus.
+		// Write bufio.Reader into a bytes.Buffer and unmarshal it so we can send it over the event bus.
 		buf := new(bytes.Buffer)
 		if _, err := buf.ReadFrom(r); err != nil {
+			return err
+		}
+
+		if err := message.UnmarshalBlock(buf, blk); err != nil {
 			return err
 		}
 


pkg/p2p/peer/processing/chainsync/sync.go
```diff
@@ -87,13 +87,18 @@ func (s *ChainSynchronizer) Synchronize(blkBuf *bytes.Buffer, peerInfo string) e
...
 		if _, err := buf.ReadFrom(r); err != nil {
 			return err
 		}
 
+		if err := message.UnmarshalBlock(buf, blk); err != nil {
+			return err
+		}
+
 		msg := message.New(topics.Block, buf)
 		s.publisher.Publish(topics.Block, msg)
```
ggoranov: That's a good catch.

Would it make sense (for sake of clarity) to have `topics.BlockWire` (accepting bytes.Buffer) that comes from the wire and `topics.Block` that is internally propagated only (accepting block.Block{} )?

 If we apply such naming convention (any messages on the wire *Wire suffixed ) we could avoid facing similar issues in future.

ggoranov: Actually, why don't we gossip buffer instead of block.Block{}?

```
func (c *Chain) propagateBlock(blk block.Block) error {
	buffer := topics.Block.ToBuffer()
	if err := message.MarshalBlock(&buffer, &blk); err != nil {
		return err
	}

	msg := message.New(topics.Block, blk)
	c.eventBus.Publish(topics.Gossip, msg)
	return nil
}
```

julesdesmit: This function never seems to be called - I think it might be legacy code from when we did not use the `Inv` messages yet.

As for `topics.BlockWire`, I don't think we need it. Wire encoded blocks are passed directly to the `Synchronizer`, who will now unmarshal it and then publish it on `topics.Block`. So no such topic would be needed.

ggoranov: It's good we have now message payload of type` interface{}`. My suggestion is to make it easier for anyone who publishes or collects a message to know if it is a buffer or generic type. This could be achieved with naming convention or at least some comments in topics package.

Currently it's very unclear what payload each topic pairs with. That might lead to more panic issues like this one.

PR #281, opened 2020-02-06
refactor RPCBus to handle interfaces instead of bytes.Buffers

### LINKED ISSUE #277: Refactor RPCBus to make use of the same `interface{}` based message passing as in the Event Bus

### LINKED ISSUE #280: Make topic encoding independent of the declaration order
The `topics.Topic` is an enum implicitly ordered due to the `iota` syntax. At the moment, the lookup of a topic, given its byte representation, is a simple and unchecked array index access. This means that, upon addition of a topic, one must make sure manually that the number expressed through its byte representation is kept consistent with the index of that topic within the enums. This is error prone and could lead to significant timesinks in case of mistakes. 

A way to enforce and check this consistency at startup would be to introduce an ordering within the `topicBuf` (the _struct_ holding the various topic's byte and string representation) and perform the consistency checks within an `init()` function

julesdesmit:
Fixes #277
Fixes #280

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/dusk-network/dusk-blockchain
 require (
 	github.com/bwesterb/go-ristretto v1.1.1
 	github.com/dusk-network/dusk-crypto v0.1.1
-	github.com/dusk-network/dusk-wallet/v2 v2.0.0
+	github.com/dusk-network/dusk-wallet/v2 v2.0.1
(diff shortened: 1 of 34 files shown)


julesdesmit:
`push` CI is currently failing due to an intermittent failure in `TestFetchDecoys`. An issue has been opened for this and will be investigated.

julesdesmit:
I will close this PR, as it was merged into #286 to be able to progress on the migration to gRPC.

PR #288, opened 2020-02-18
Extend graphql service to satisfy dusk-explorer back-end needs

### LINKED ISSUE #276: Migrate webserver/events package from dusk-explorer into dusk-blockchain repo
1. Run a new configurable HTTP service to handle websocket connections
2. Reuse webserver/events  from dusk-explorer
3. To be disabled by default
4. Provide blocks count value

ggoranov:
Fixes #276 #289  #290

### DIFF
--- a/pkg/gql/notifications/broker.go
+++ b/pkg/gql/notifications/broker.go
@@ -1,0 +1,205 @@
+package notifications
+
+import (
+	"time"
+
+	"container/list"
+
+	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/topics"
+	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/eventbus"
+	"github.com/dusk-network/dusk-wallet/v2/block"
+	logger "github.com/sirupsen/logrus"
+)
+
+const (
+	// Write deadline. After a write has timed out, the websocket state is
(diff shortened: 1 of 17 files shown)


ggoranov:
CI fails due to  `--- FAIL: TestFetchDecoys (0.00s) drivers_test.go:820: incorrect amount of hits - 6/7`
I suspect `pull/284` will fix this. Until it's merged, this PR can be reviewed.

pkg/gql/notifications/broker.go
```diff
@@ -0,0 +1,205 @@
...
+
+	log.Tracef("handleConn %s", conn.RemoteAddr().String())
+
+	b.reap()
+
+	if b.clients.Len() >= int(b.maxClientsCount) {
+		log.Errorf("Too many connections %d", b.clients.Len())
+
+		// Free a slot by removing the oldest connection
+		e := b.clients.Front()
```
autholykos: Maybe it would be a bit of over-engineering, however, in order to prevent force-dropping connections, oldest connections should be dropped only when at full capacity, after reaching a maturity threshold (connections older than X minutes). Additionally, incoming connections, when at full capacity, should enqueue with a random delay to prevent spamming. This would be good as a separate issue

autholykos: The limiter middleware kinda takes care of this already. Please ignore

autholykos:
LGTM

PR #297, opened 2020-02-28
Peer reader handles invalid frames

### LINKED ISSUE #294: Panic observed if a malformed wire message is sent
Node panics when frame size field has values in the interval  [5,7] due to lack of len check in `checksum.Extract` 

To reproduce:

```go
        frame := new(bytes.Buffer)
	// Add length bytes
	if err := encoding.WriteUint64LE(frame, 7); err != nil {
		return err
	}

        // Add correct magic value
	magic := protocol.TestNet
	mBuf := magic.ToBuffer()
	if _, err := frame.Write(mBuf.Bytes()); err != nil {
		return err
	}
       
       // Checksum with 3 bytes only
       if _, err := frame.Write([]byte{0, 0, 0}); err != nil {
			t.Error(err)
       }
        // malformed payload ( 3 bytes-only checksum, 0 message )
	_, err := c.Write(frame.Bytes())
```

```go
panic: runtime error: slice bounds out of range [:4] with capacity 3

goroutine 37 [running]:
github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/checksum.Extract(...)
       dusk-blockchain/pkg/p2p/wire/checksum/checksum.go:18
github.com/dusk-network/dusk-blockchain/pkg/p2p/peer.(*Reader).ReadLoop(0xc0003002e0)
      dusk-blockchain/pkg/p2p/peer/peer.go:237 +0x6cf
created by main.(*Server).OnAccept
       dusk-blockchain/launch/testnet/server.go:147 +0x553
```

ggoranov:
Fix #294 , #303, #304

### DIFF
--- a/pkg/p2p/peer/peer.go
+++ b/pkg/p2p/peer/peer.go
@@ -242,7 +242,7 @@ func (p *Reader) readLoop() {
 		// Refresh the read deadline
 		err := p.Conn.SetReadDeadline(time.Now().Add(readWriteTimeout))
 		if err != nil {
-			l.WithError(err).Warnf("write timeout err: %v", err)
+			l.WithError(err).Warnf("error setting read timeout")
 		}
 
 		b, err := p.ReadMessage()
--- a/pkg/p2p/peer/peer.go
+++ b/pkg/p2p/peer/peer.go
@@ -213,6 +213,20 @@ func (w *Writer) writeLoop(writeQueueChan <-chan *bytes.Buffer, exitChan chan st
 // is reached. Should be called in a go-routine, after a successful handshake with
 // a peer. Eventual duplicated messages are silently discarded.
 func (p *Reader) ReadLoop() {
+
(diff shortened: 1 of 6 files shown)


pkg/p2p/peer/peer.go
```diff
@@ -226,7 +240,10 @@ func (p *Reader) ReadLoop() {
 
 	for {
 		// Refresh the read deadline
-		p.Conn.SetReadDeadline(time.Now().Add(readWriteTimeout))
+		err := p.Conn.SetReadDeadline(time.Now().Add(readWriteTimeout))
+		if err != nil {
+			l.WithError(err).Warnf("write timeout err: %v", err)
```
julesdesmit: In accordance with the style guide, please don't format the error into the `msg` field - having the `WithError` is enough.

ggoranov: Done. 
That reminds that we'd need to release a coding guidelines doc that defines such rules/advises.

PR #343, opened 2020-03-13
Keep track of republished candidates separately from the queue

### LINKED ISSUE #341: Candidate is not found intermittently, when finalizing a consensus round

julesdesmit:
Fixes #341

### DIFF
--- a/pkg/core/candidate/broker.go
+++ b/pkg/core/candidate/broker.go
@@ -59,7 +59,7 @@ func NewBroker(broker eventbus.Broker, rpcBus *rpcbus.RPCBus) *Broker {
 		bestScoreChan:     bestScoreChan,
 	}
 
-	broker.Subscribe(topics.ValidCandidateHash, eventbus.NewCallbackListener(b.addValidHash))
+	broker.Subscribe(topics.ValidCandidateHash, eventbus.NewCallbackListener(b.AddValidHash))
 	b.republisher = republisher.New(broker, topics.Candidate, Validate, b.Validate)
 	return b
 }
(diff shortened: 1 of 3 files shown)


pkg/core/candidate/broker.go
```diff
@@ -171,12 +174,13 @@ func (b *Broker) requestCandidate(hash []byte) (message.Candidate, error) {
...
-	b.lock.RLock()
-	defer b.lock.RUnlock()
-	if _, ok := b.queue[string(cm.Block.Header.Hash)]; ok {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+	if _, ok := b.republished[string(cm.Block.Header.Hash)]; ok {
 		return republisher.DuplicatePayloadError
 	}
 
+	b.republished[string(cm.Block.Header.Hash)] = struct{}{}
```
autholykos: any reason why this is not simply a `delete(b.republished, string(cm.Block.Header.Hash))`?

pkg/core/candidate/broker.go
```diff
@@ -195,3 +199,11 @@ func (b *Broker) clearEligibleHashes() {
 	}
 	b.lock.Unlock()
 }
+
+func (b *Broker) clearRepublishedHashes() {
+	b.lock.Lock()
```
autholykos: according to our guidelines, unexported functions should delegate locking to the caller, so to prevent deadlocks and excessive locking

pkg/core/candidate/broker.go
julesdesmit: Why would I delete something that's not there in the first place?

autholykos: ah yes, completely misread the code. Was mislead by the fact that a `map` is used instead of a `hashset`

PR #345, opened 2020-03-16
Create gRPC client for interaction with the Rusk server

### LINKED ISSUE #344: Create a gRPC client for interaction with the Rusk server

julesdesmit:
Fixes #344


julesdesmit:
This should be ready to review on its own now. There are still a couple things to work out on other ends right now though, the main items being Phoenix integration, and some updates to the protobuf structures.

pkg/rpc/client.go
```diff
@@ -0,0 +1,60 @@
...
+
+// InitRuskClient opens the connection with the Rusk gRPC server, and
+// launches a goroutine which listens for RPCBus calls that concern
+// the Rusk server.
+//
+// As the Rusk server is a fundamental part of the node functionality,
+// this function will panic if the connection can not be established
+// successfully.
+func InitRuskClient(address string, rpcBus *rpcbus.RPCBus) {
+	conn, err := grpc.Dial(address, grpc.WithInsecure(), grpc.WithBlock())
```
ggoranov: This will block infinitely if Rust server is down/not responding. We can use dial timeout here and then panic?

julesdesmit: Good point

ggoranov: At closing the node or restarting the conn, we could need conn.Close()

autholykos:
blocked by #380

PR #346, opened 2020-03-18
Harness refactor

ggoranov:
Fixes https://github.com/dusk-network/dusk-blockchain/issues/287

### DIFF
--- a/harness/tests/localnet_test.go
+++ b/harness/tests/localnet_test.go
@@ -1,18 +1,16 @@
 package tests
 
 import (
+	"encoding/hex"
 	"flag"
-	"fmt"
 	"io/ioutil"
 	"os"
-	"strconv"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


harness/tests/localnet_test.go
```diff
@@ -74,145 +70,78 @@ func TestMain(m *testing.M) {
...
+		if i == 0 {
+			blockhash = bh
 		}
 	}
 }
 
 // TestCatchup tests that a node which falls behind during consensus
 // will properly catch up and re-join the consensus execution trace.
 func TestCatchup(t *testing.T) {
+
```
ggoranov: @julesdesmit Let me know if the refactored TestCatchup works for you as I tried to avoid the need of "PublishTopic" rpc call. If not, I could restore the initial idea test and add `PublishTopic` to gRPC node service.

julesdesmit: Yep, both tests are passing without issue on my end.

PR #362, opened 2020-03-30
move testnet and kadcast to cmd dir and fix all related

devtomnl:
* update Makefile
* update travis build
* add scripts/build.go to be able to build packages with go
--
* pending: scripts shall be useful or removed ?


scripts/build.go
```diff
@@ -0,0 +1,61 @@
+// +build none
+
+package main
+
+import (
+	"fmt"
+	"log"
```
autholykos: Any reason why using vanilla golang `log` instead of `logrus` in the main?

autholykos:
> pending: scripts shall be useful or removed?
The script's only use would be to startup the blindbid process. Most likely, this could be substituted with a docker manifest once wiring the Rusk process

cmd/dusk/server.go
```diff
@@ -176,7 +166,7 @@ func (s *Server) OnConnection(conn net.Conn, addr string) {
 // Close the chain and the connections created through the RPC bus
 func (s *Server) Close() {
 	// TODO: disconnect peers
-	s.chain.Close()
+	_ = s.chain.Close()
```
ggoranov: For this and other lint errors, we can have a separate issue that fixes all lint issues but also enabled CI lint checks.

scripts/build.go
```diff
@@ -0,0 +1,61 @@
...
+	default:
+		log.Fatal("cmd not found", os.Args[1])
+	}
+}
+
+func install() {
+
+	argsList := append([]string{"list"}, []string{"./..."}...)
+
+	cmd := exec.Command(filepath.Join(runtime.GOROOT(), "bin", "go"), argsList...)
```
ggoranov: Do we have a README.md that gives examples of a proper use? (ignore if this has been already raised)

cmd/dusk/action.go
```diff
@@ -3,35 +3,51 @@ package main
...
+	if arguments := ctx.Args(); len(arguments) > 0 {
+		return fmt.Errorf("failed to read command argument: %q", arguments[0])
+	}
+
+	if datadir = ctx.GlobalString(DataDirFlag.Name); datadir != "" {
+		datadir = "~/./dusk"
+		//TODO:
+	}
+
+	if config = ctx.GlobalString(ConfigFlag.Name); config != "" {
```
ggoranov: Doesn't this duplicate  `loadFlags` we already support (--config)?

https://github.com/dusk-network/dusk-blockchain/blob/ffb61f2bf7580bac17c841ce75dbe95f5de77ef3/pkg/config/loader.go#L175

cmd/dusk/action.go
```diff
@@ -111,7 +131,7 @@ func setupProfiles(r *rpcbus.RPCBus) *diagnostics.ProfileSet {
 
 			p := diagnostics.NewProfile(i.Name, i.Interval, i.Duration, i.Start)
 			if err := s.Spawn(p); err != nil {
-				log.Panicf("Profiling task error: %s", err.Error())
+				log.WithError(err).Panic("Profiling task error")
```
ggoranov: All these improper log entries can be changed  by a separate issue where we translate `log.*f(..., err)` to `log.WithError(err)` as per the guideline.

cmd/dusk/action.go
```diff
@@ -3,35 +3,51 @@ package main
...
 
-func main() {
+func action(ctx *cli.Context) error {
+
+	// check arguments
+	if arguments := ctx.Args(); len(arguments) > 0 {
+		return fmt.Errorf("failed to read command argument: %q", arguments[0])
+	}
+
+	if datadir = ctx.GlobalString(DataDirFlag.Name); datadir != "" {
```
ggoranov: Would you confirm that test harness is compatible with CLI changes?

scripts/build.go
devtomnl: this is not the main, its the build script. If we try to use logrus, nothing will be printed.

scripts/build.go
devtomnl: Good point, I'll update the readme.

cmd/dusk/action.go
devtomnl: Agree, I'll raise a ticket to continue the changes. But keeping it in a new file does not make sense, so I updated already. Are you suggesting to revert it ?

cmd/dusk/action.go
devtomnl: I did not test that, I'll check.

cmd/dusk/server.go
devtomnl: sure, there is already #366 opened to address this. Leaving it unfixed in this PR seemed a bit weird. So I suggest we leave it fixed instead of undoing it. Would you agree ?

ggoranov: Sure.

cmd/dusk/action.go
ggoranov: No need to revert, just a suggestion for a new issue.

cmd/dusk/action.go
devtomnl: it seems to work fine

scripts/build.go
devtomnl: done

cmd/dusk/action.go
devtomnl: good point. However when trying to re-use that method I stumbled in few issues: 1. `loadFlags` its not exported, so its not possible to use it in the main. 2. When making it exported and trying to re-use its logic, I started to face several other issues like: 
`INFO[0000] initializing node...                          app=dusk prefix=main
Dusk node flag redefined: general.network
ERRO[0000] Application panic                             app=dusk error="Dusk node flag redefined: general.network" prefix=main`
3. The code was already like this before, but the config was hardcoded. Instead, I improved a bit by allowing it to be configurable. 

I think we should refactor this flags in another ticket to be raised.

devtomnl: raised #367

PR #391, opened 2020-04-08
 add test-harness to travis ci #368 

devtomnl:
* add test-harness-ci and get-blindbid to Makefile
* add Harness step to travis build
* update localnet_test.go to accept NETWORK_SIZE as env var or use default when not set
* update travis build names


cmd/dusk/monitor.go
```diff
@@ -4,13 +4,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
julesdesmit, line 7: Seems like logrus is imported twice now. Let's stick to just one, to keep the code clean :)

devtomnl: done

Issue #301, opened 2020-03-02
Block structure should be updated in accordance with Rusk


julesdesmit:
The `Block` structure should have the `Tx` field updated according to the Tx structures defined in the dusk-network/dusk-protobuf.

This issue includes includes removing dependencies to `data/transactions` package which uses legacy transaction creation

julesdesmit:
Fixed via #450

Issue #300, opened 2020-03-02
Verification package should be deprecated in favor of gRPC calls to Phoenix server


julesdesmit:
The Phoenix server will currently be the 'rusk' host environment.

julesdesmit:
Fixed via #450

Issue #299, opened 2020-03-02
Re-generate the genesis block in accordance with Phoenix


julesdesmit:

autholykos:
Duplicate of #417

Issue #298, opened 2020-03-02
Replace current UTXO model for Phoenix


julesdesmit:
This entails that the wallet no longer does its own generation of transactions, but instead reaches out to a 'phoenix server' process (over gRPC), to generate the data. The `Transactor` should be deprecated, and `dusk-wallet` could become a gRPC client.

autholykos:
closing as it is a duplicate of #411

PR #398, opened 2020-04-10
Switch from legacy wallet to phoenix-wallet

### LINKED ISSUE #363: Substitute legacy wallet with Rusk RPCs
The following changes should happen in the repository in order to fully support Phoenix as a transaction model accessed through rusk process:

## Database
- Blocks should now store phoenix transactions (generated protobuf structures)
- Locktimes and related functions no longer need to be recorded, as locking up of stakes and bids will happen through a contract
- KeyImages and related functions will no longer need to be recorded, as the mempool (using a phoenix RPC call) and the state transition function should guard against double spends
- FetchDecoys will no longer be needed

## Chain
- `restoreConsensusData` can be removed, as consensus data can be fetched from global storage.
- No longer needs to hold the bidlist and provisioners. Instead, when updating the round, it should get these from the global storage.
- `addConsensusNodes` and all related functionality can be removed
- `AcceptBlock` should make a call to ExecuteStateTransition (RPCBus topic already exists), right after `verifiers.CheckBlock`
- `removeXXX` functions for provisioners and block generators can be removed

## Verifiers
- Transaction file can be removed entirely, or can be whittled down to one single function which calls a phoenix RPC call to verify a transaction. NOTE that this call should not verify the proof - it should simply do some correctness checks on the tx values. The proof is verified later, in the smart contract execution.
- Tx related functions in block file can be removed. This includes `CheckMultiCoinbases` and `CheckTx` calls.

## Mempool
- Depending on transaction verification overhaul, either call VerifyTransaction directly over the RPCBus, or use the method in the `verifiers/transaction.go` file.
- `CheckTxDoubleSpent` can be removed, as VerifyTransaction should check for double spends by scanning for the nullifier in the nullifier set
- Should get a method which takes a slice of phoenix transactions, that have failed execution during block generation. This way, obsolete transactions can be removed each round.

## Transactor
- All dusk-wallet related logic needs to be replaced for phoenix RPC calls where appropriate, as all of the logic will reside in a rust process.

This list may be incomplete for the time being, but it should cover most of the affected areas.

autholykos:
Resolves #363 #301 #298 #299 #300


autholykos:
depends on dusk-network/dusk-protobuf#32

autholykos:
This PR is WAY too big, is dependant on a full migration to a rusk wallet. Closing in favor of a set of incremental PRs

PR #403, opened 2020-04-11
Wallet/feature 400

autholykos:


autholykos:
> It looks like the `block` package was not migrated yet.

I understand now what you meant with migrating `block` package. I see Travis failing because of that. However, I don't understand how this is happening since 1) on my machine everything seems to run just fine (with the same `golangci.yml` configuration and 2) the offending package `pkg/core/data/block` is regularly found in the codebase

julesdesmit:
> the offending package pkg/core/data/block is regularly found in the codebase

It seems not, for me anyways.
Screenshot 2020-04-14 at 08 36 05
Screenshot 2020-04-14 at 08 35 30

autholykos:
Screenshot 2020-04-14 at 08 38 14

Weird. There is nothing in my `.gitignore` to prevent `pkg/core/data/block` to be regularly added and committed

julesdesmit:
@autholykos I think I might know what it is - I had struggled with something similar when I was migrating the `block` out of this repo. In `.gitignore`, there's a line `*LOCK` which I believe might be the offender here. It's supposed to keep leveldb files out of the repo but it affects the block package inadvertently.

autholykos:
Indeed that was the problem. A `git status --ignored` highlighted that indeed the package was ignored. I force added `block` package and it should be OK now. I will also modify the `*LOCK` entry on `.gitignore`, so to just pick the leveldb files

autholykos:
actually, a better way is to change locally the `.git/config` to let git be case-sensitive:
```
...
[core]
    ignorecase = false
...
```
```

Issue #405, opened 2020-04-14
cross compiling dusk to 386 fails on transactions.go:106


devtomnl:
When trying to enable dusk cross compiling to linux/386 and darwin/386, I stumbled into the following err:

```
# github.com/dusk-network/dusk-blockchain/pkg/core/database/heavy
../../pkg/core/database/heavy/transactions.go:106:16: constant 4294967295 overflows int
2020/04/14 11:09:42 Failed to cross compile package: exit status 2.
make: *** [Makefile:41: dusk-linux] Error 1
```

related to: #390

autholykos:
depends on #363

devtomnl:
found a fix for it. The trick is convert the len to uint32 before the comparison:

```
if uint32(len(b.Txs)) > math.MaxUint32 {
	return errors.New("too many transactions")
}
```

devtomnl:
It turns out that link complains about the statement:
```
pkg/core/database/heavy/transactions.go:107:5: SA4003: no value of type uint32 is greater than math.MaxUint32 (staticcheck)

	if uint32(len(b.Txs)) > math.MaxUint32 {

	   ^
```

so, updating it from uint32 to uint64 makes the trick and.

Issue #407, opened 2020-04-14
dusk cross compiling to windows (amd64/386) fails with undefined: syscall.Mkfifo


devtomnl:
```
# github.com/dusk-network/dusk-zkproof
/go/pkg/mod/github.com/dusk-network/dusk-zkproof@v0.0.0-20190727103229-8b0c008561ee/pipe.go:21:2: undefined: syscall.Mkfifo
2020/04/14 11:44:20 Failed to cross compile package: exit status 2.
make: *** [Makefile:47: dusk-windows] Error 1
```

related to: #390

PR #408, opened 2020-04-14
enable cross compiling of dusk-blochain cmds #390 

devtomnl:
While working on this, I have found two issues with the current codebase that would have to be fixed in order for us to have all the pending cross compiling: #405 and  #407

##edit
closes: #405


autholykos:
depends on #363

Issue #411, opened 2020-04-16
Modify Transactor to use rusk RPC where appropriate


autholykos:
Transactor is a thin layer that reroute (cli- web-)wallet requests toward the appropriate endpoint. This issue is about phasing out the `wallet.Wallet` (reliant on UTXO and bulletproofs) and use rusk GRPC services instead (which use PLONK, phoenix transaction model and smart-contracts). In particular, this means: 
- remove dependency on `data/wallet`
- remove the decoy creation/handling (since they are specific to ringsignature)
- consider fetching inputs directly from rusk

julesdesmit:
@devtomnl  

## Transactor commands, and what they should do

We can leave wallet DB stuff unimplemented for now, as it might take quite some work to get the DB itself adjusted to work with Phoenix as well. I'm just leaving them here for reference

### GetTxHistory

- Retrieve the transaction records from the database. Sort them based on date, and return the most recent 10.

### CreateWallet

- Make a call to Rusk, method `GenerateSecretKey`. It takes some bytes as a seed, but this can be randomly generated in the function itself.
- Once the secret key is received, the Transactor should set it on its struct.
- The secret key needs to be stored on disk, in an encrypted file. In the past, we had an AES-encrypted wallet.dat file which needed a password to decrypt.
- Make another call to rusk, method `Keys`. This returns the view key and public key.
- Return the public key.

### LoadWallet

- Should attempt to decrypt the on-disk file. On success, sets the private key on the transactor struct.
- Make a call to rusk, method `Keys`. Return the public key.

### CreateFromSeed

- Same as CreateWallet, but the seed can be passed as an argument.

### ClearWalletDatabase

- Removes the on-disk wallet database.

### CallContract

- If private key is set, fetch inputs from the wallet db until we have enough to cover the transaction (amount inputted >= amount to send).
- Make a call to rusk, method `NewTransaction`. (NOTE: `NewTransaction` currently does not yet support adding calldata to transactions. You can leave a TODO there for now.)
- Send the received transaction to the node's mempool (this can go over the event bus, or the rpc bus).
- Return the tx hash.

### Transfer

- If private key is set, fetch inputs from the wallet db until we have enough to cover the transaction (amount inputted >= amount to send).
- Make a call to rusk, method `NewTransaction`.
- Send the received transaction to the node's mempool (this can go over the event bus, or the rpc bus).
- Return the tx hash.

### Bid

[...]

PR #450, opened 2020-04-24
[WIP] Modify Transactor to use rusk RPC where appropriate #411

devtomnl:


autholykos:
rather than using RPC, the Transactor should have the proper interface defined in `pkg/code/data/transactions/provider.go` injected and use that one

cmd/dusk/server.go
```diff
@@ -44,13 +46,14 @@ type Server struct {
...
 // component and performs a DB sanity check
-func LaunchChain(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, counter *chainsync.Counter, srv *grpc.Server) (chain.Loader, error) {
+func LaunchChain(ctx context.Context, proxy *transactions.Proxy, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, counter *chainsync.Counter, srv *grpc.Server) (chain.Loader, error) {
 	// creating and firing up the chain process
 	genesis := cfg.DecodeGenesis()
 	_, db := heavy.CreateDBConnection()
 	l := chain.NewDBLoader(db, genesis)
 
-	chainProcess, err := chain.New(eventBus, rpcBus, counter, l, l, srv)
+	// TODO: inject the proper interface
```
julesdesmit, line 55: Is `proxy.Executor` not the right interface? Just wanna make sure we don't leave and TODOs lying around that aren't necessary :)

pkg/core/candidate/store_test.go
```diff
@@ -73,14 +73,17 @@ func TestRequestCandidate(t *testing.T) {
 				return
 			}
 
-			for i, tx := range genesis.Txs {
-				otherTx := c.Block.Txs[i]
-
-				if !tx.Equals(otherTx) {
-					errChan <- fmt.Errorf("Candidate Block Tx #%d is different from genesis block Tx", i)
-					return
+			// TODO: add equals methods for phoenix/RUSK transactions
```
julesdesmit, line 76: This method should exist now, so we can reinstate this test

pkg/core/chain/loader.go
```diff
@@ -52,14 +52,15 @@ func (l *DBLoader) CheckBlock(prevBlock block.Block, blk block.Block) error {
...
-		tx, ok := merklePayload.(transactions.Transaction)
+	for _, merklePayload := range blk.Txs {
+		_, ok := merklePayload.(transactions.ContractCall)
 		if !ok {
 			return errors.New("tx does not implement the transaction interface")
 		}
-		if err := verifiers.CheckTx(l.db, uint64(i), uint64(blk.Header.Timestamp), tx); err != nil {
-			return err
-		}
+		// TODO: should be a call to rusk method `VerifyTransaction`
```
julesdesmit, line 60: This can be added once the `transactor-provider` branch is merged in. We can defer it to another issue though, if you'd like.

pkg/core/candidate/store_test.go
julesdesmit: Can also be deferred to another issue.

pkg/rpc/client.go
```diff
@@ -64,7 +70,7 @@ func (c *Client) listen() {
 	for {
```
julesdesmit, line 70: Since we seem to be passing the rpc clients around with interfaces, we can probably get rid of the listening goroutine.

autholykos: Agreed, the indirection looks unnecessary

PR #467, opened 2020-05-05
Transaction provider

autholykos:
This PR introduces the following interfaces:
- Verifier: abstracts away the GRPC calls related to transaction verification
- Provider: abstracts away the GRPC calls related to Transaction creation
- Executor: abstracts away the GRPC calls related to Transaction execution
- Provisioner: abstracts away the GRPC calls related to Provisioner operations 
- BlockGenerator: abstracts away the GRPC calls related to Block Generator operations
- Consensus: abstracts away the GRPC calls related to BidList and Committee calculation

This layer of indirection allows surgical injection of just the right operations into the various node components, which presents also the added benefit of ease of testing and separation of concerns.

The PR includes all needed serializations to and from `rusk` protobuf structures as well as new transaction objects


pkg/core/chain/chain.go
```diff
@@ -200,11 +206,17 @@ func (c *Chain) Listen() {
 			c.provideLastCertificate(r)
 		case r := <-c.getRoundResultsChan:
 			c.provideRoundResults(r)
+		case msg := <-c.acceptBlockChan:
+			if err := c.onAcceptBlock(ctx, msg); err != nil {
```
julesdesmit, line 210: I'd advise to keep the callback over the channel. The reason for this being that the `Synchronizer` relies on `AcceptBlock` being blocking, otherwise it causes a race condition between when a block is being accepted, and when a new one is downloaded and attempted to be passed down. Since we determine whether or not to pass a block down based on the `prevBlock.Height`, it can result in blocks being dropped inadvertently.

pkg/core/consensus/user/sortition.go
```diff
@@ -5,12 +5,16 @@ import (
...
-	"github.com/dusk-network/dusk-blockchain/pkg/core/data/wallet"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/sortedset"
 	"github.com/dusk-network/dusk-crypto/hash"
 	log "github.com/sirupsen/logrus"
 )
 
+// DUSK is one whole unit of DUSK. This is duplicated from wallet since
+// otherwise we get into an import cycle including the transactions and users
+// packages
+const DUSK = uint64(100000000)
```
julesdesmit, line 16: Could we avoid cyclic imports and duplication by just putting this in the `config` package?

pkg/core/chain/chain.go
autholykos: problem with that is that we now need a `context.Context` which should be injected when the call is done. It is bad practice to save it in the chain and I am not sure how we can add it to the method signature directly within the callback

pkg/core/consensus/user/sortition.go
autholykos: yes, this was a quick hack. Did not give it that much thought

pkg/core/chain/chain.go
autholykos: mmm apparently, the advice to not store context can be relaxed a bit (see: https://github.com/golang/go/issues/22602). I will revert the change and keep the context as a field within the `Chain`

pkg/core/consensus/user/sortition.go
autholykos: will open a separate ticket for it

Issue #456, opened 2020-05-04
key.Key should hold Phoenix keys


julesdesmit:

autholykos:
`StealthAddress` can be removed as well as `PublicSpend`. Everything else should be the SK, the PK and the VK returned by Rusk

julesdesmit:
Fixed via #472

Issue #470, opened 2020-05-06
Inject provider interface instead of Rusk in all node components


autholykos:

autholykos:
The transactor has been adjusted in #472

PR #472, opened 2020-05-06
Transactor provider

autholykos:
part of #470

### DIFF
--- a/pkg/core/data/transactions/fields.go
+++ b/pkg/core/data/transactions/fields.go
@@ -288,6 +288,11 @@ type SecretKey struct {
 	B *Scalar `json:"b"`
 }
 
+// IsEmpty is used when passing the SecretKey by value
+func (s SecretKey) IsEmpty() bool {
+	return s.A == nil && s.B == nil
+}
+
 // MSecretKey copy the USecretKey from rusk to transactions datastruct
 func MSecretKey(r *rusk.SecretKey, n *SecretKey) {
 	r.A = new(rusk.Scalar)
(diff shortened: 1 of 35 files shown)


autholykos:
fixes #472

pkg/core/data/transactions/fields.go
```diff
@@ -288,6 +288,11 @@ type SecretKey struct {
 	B *Scalar `json:"b"`
 }
 
+// IsEmpty is used when passing the SecretKey by value
+func (s SecretKey) IsEmpty() bool {
+	return s.A == nil && s.B == nil
+}
+
 // MSecretKey copy the USecretKey from rusk to transactions datastruct
```
julesdesmit, line 296: This comment is the same as the one for `USecretKey`, so it's a bit hard to know which one is going in which direction unless you look at the logic

pkg/core/data/transactions/fields.go
```diff
@@ -288,6 +288,11 @@ type SecretKey struct {
...
 func MSecretKey(r *rusk.SecretKey, n *SecretKey) {
 	r.A = new(rusk.Scalar)
```
julesdesmit, line 298: Should we have the `IsEmpty` check here as well, like in `MPublicKey`?

pkg/core/mempool/mempool.go
```diff
@@ -65,37 +63,43 @@ type Mempool struct {
 	latestBlockTimestamp int64
 
 	eventBus *eventbus.EventBus
-	db       database.DB
+	// TODO: do we still need a DB in the mempool?
```
julesdesmit, line 66: Nope, as discussed, Rusk will check the DB for double-spends when verifying a mempool tx.

pkg/core/data/transactions/fields.go
autholykos: good point, will fix

pkg/core/data/transactions/fields.go
autholykos: In theory sounds good. However, the `SecretKey` is normally required. I will fix it but we might want to add some verification by the caller

pkg/core/mempool/mempool.go
autholykos: Removed the comment

autholykos:
Fixes #475

autholykos:
Solves #479

autholykos:
Fixes #456

PR #490, opened 2020-05-11
Provide the block generator with the latest winning committee

### LINKED ISSUE #478: Make the BlockGenerator able to reconstruct the Committee of the previous block to create a correct Coinbase tx
The `BlockGenerator` needs the know the participants of Provisioners' committee that finalized the previous block in order to reward them through the `DistributeTransaction`. Unfortunately, the block certificate only carries the bit representation of those Provisioners which needs the full *ordered* set of Provisioners to identify those belonging in the committee.

julesdesmit:
Fixes #478


pkg/core/consensus/candidate/blockgenerator.go
```diff
@@ -103,7 +103,13 @@ func (sf ScoreFactory) Create(sender []byte, round uint64, step uint8) consensus
 func (bg *Generator) Collect(e consensus.InternalPacket) error {
 	sev := e.(message.ScoreProposal)
 
-	blk, err := bg.Generate(sev)
+	resp, err := bg.rpcBus.Call(topics.GetLastCommittee, rpbus.EmptyRequest(), 5*time.Second)
+	if err != nil {
+		return err
+	}
+	keys := resp.([][]byte)
```
autholykos, line 110: This will throw an error. The Chain responds to `GetLastCommittee` requests with a `bytes.Buffer` carrying the wire-encoded BLS keys of the committee (which I think it is unnecessary). Either change `chain.provideLastCommittee` to return a `[][]byte` or change this line to cast to a `bytes.Buffer`

autholykos: My preference goes for the former of the 2 options.

pkg/core/consensus/agreement/handler_test.go
```diff
@@ -29,6 +30,27 @@ func TestVoteVerification(t *testing.T) {
...
+	// Ensure voterKeys only contains keys from `keys`
+	for _, key := range voterKeys {
+		found := false
+		for _, k := range keys {
+			if bytes.Equal(k.BLSPubKeyBytes, key) {
+				found = true
+			}
+		}
+
+		assert.True(t, found)
```
autholykos, line 50: use `require` instead of `assert`. Aliase it with assert at the beginning of the test

pkg/core/chain/chain.go
```diff
@@ -526,6 +534,26 @@ func (c *Chain) provideLastCertificate(r rpcbus.Request) {
...
+	}
+
+	for _, key := range c.lastCommittee {
+		if err := encoding.WriteBLS(buf, key); err != nil {
+			r.RespChan <- rpcbus.NewResponse(bytes.Buffer{}, err)
+		}
+	}
+
+	r.RespChan <- rpcbus.NewResponse(*buf, nil)
+}
```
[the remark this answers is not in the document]
julesdesmit: You're entirely right - I totally missed that. Will change it

Issue #476, opened 2020-05-07
Block generator `K` and `D` should be put into storage when a bid transaction is made


julesdesmit:

autholykos:
We should also add the ed25519 PublicKey used to identify the `Bid` transaction

autholykos:
Closed via #510

PR #510, opened 2020-05-13
Block generator `K` and `D` should be put into storage when a bid transaction is made #476

devtomnl:


autholykos:
The tests are failing due to #531. Merging since the problem will be fixed in another PR

Issue #496, opened 2020-05-11
Get tests back in shape


autholykos:

autholykos:
Tests are now passing on `rusk-ipc`

PR #514, opened 2020-05-13
Regenerate genesis block, and fix related tests

### LINKED ISSUE #417: Regenerate Genesis block
When migrating to rusk, the genesis block needs to be regenerated since the staking and bid transactions in the current genesis block will no longer be valid

julesdesmit:
Fixes #417 

Related to #496


autholykos:
LGTM, just nitpinching on some tests and there is a bit of weirdness in instantiating the `DistributeTransaction` (I don't like when calling a `MockXXX` function to instantiate a real struct), but I will tackle them in another PR

PR #525, opened 2020-05-14
Reword TxRecord to be in line with the current transaction model

autholykos:


autholykos:
This PR introduces a regression on `gql/query`

PR #529, opened 2020-05-15
Implement Prometheus endpoints enabled by cli config #528

devtomnl:


devtomnl:
this PR is no longer required since the scope has changed.

PR #532, opened 2020-05-16
Solve republishing-related race conditions

### LINKED ISSUE #531: Fix race condition between the Republisher and the Agreement accumulator
Travis complains about a novel race condition that is seemingly caused by the `Republisher` trying to marshal an `Agreement` message concurrently with the collector's verification causing the stack trace below.

The best solution would be to remove the concurrent republishing and perform it either before (i.e. prioritizing latency over bandwidth) or after validating the message (i.e. prioritizing bandwidth over latency).

We need to investigate whether other messages are affected by the republisher's concurrency
 
```
Previous write at 0x00c0007f2b60 by goroutine 35:
  github.com/dusk-network/bn256.(*curvePoint).MakeAffine()
      /Users/julesdesmit/go/pkg/mod/github.com/dusk-network/bn256@v0.5.1-lattices/curve.go:234 +0x8e0
  github.com/dusk-network/bn256.(*G1).Marshal()
      /Users/julesdesmit/go/pkg/mod/github.com/dusk-network/bn256@v0.5.1-lattices/bn256.go:117 +0x8c
  github.com/dusk-network/bn256.(*G1).Compress()
      /Users/julesdesmit/go/pkg/mod/github.com/dusk-network/bn256@v0.5.1-lattices/bn256.go:134 +0x74
  github.com/dusk-network/dusk-crypto/bls.(*Signature).Compress()
      /Users/julesdesmit/go/pkg/mod/github.com/dusk-network/dusk-crypto@v0.1.1/bls/bls.go:285 +0x6f
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.MarshalStepVotes()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/agreement.go:364 +0x26e
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.MarshalVotes()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/agreement.go:342 +0x1a2
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.MarshalAgreement()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/agreement.go:212 +0x3b4
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.marshalMessage()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/message.go:227 +0x8d8
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.marshal()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/message.go:200 +0x357
  github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message.Marshal()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/message.go:165 +0x489
  github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/republisher.(*Republisher).Republish()
      /Users/julesdesmit/go/src/github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/republisher/republisher.go:97 +0x3cb
```

autholykos:
Fixes #531

### DIFF
--- a/pkg/core/candidate/broker.go
+++ b/pkg/core/candidate/broker.go
@@ -64,6 +64,7 @@ func NewBroker(broker eventbus.Broker, rpcBus *rpcbus.RPCBus) *Broker {
 
 	broker.Subscribe(topics.ValidCandidateHash, eventbus.NewCallbackListener(b.AddValidHash))
 	b.republisher = republisher.New(broker, topics.Candidate, Validate, b.Validate)
+	b.republisher.Go()
 	return b
 }
 
--- a/pkg/core/candidate/broker.go
+++ b/pkg/core/candidate/broker.go
@@ -64,7 +64,6 @@ func NewBroker(broker eventbus.Broker, rpcBus *rpcbus.RPCBus) *Broker {
 
 	broker.Subscribe(topics.ValidCandidateHash, eventbus.NewCallbackListener(b.AddValidHash))
 	b.republisher = republisher.New(broker, topics.Candidate, Validate, b.Validate)
-	b.republisher.Go()
 	return b
 }
 
--- a/pkg/core/consensus/agreement/accumulator.go
+++ b/pkg/core/consensus/agreement/accumulator.go
@@ -38,7 +38,7 @@ func newAccumulator(handler Handler, workerAmount int) *Accumulator {
 func (a *Accumulator) Process(ev message.Agreement) {
 	defer func() {
 		// we recover from panic in case of a late Process call which would attempt to write to the closed verificationChan
-		// the alternative would be to never close the verificationChan and either use a multitude of channels to  stop the workers or a shared boolean set in the Accumulator.Stop
+		// the alternative would be to never close the verificationChan and either use a multitude of channels to stop the workers or a shared boolean set in the Accumulator.Stop
 		if r := recover(); r != nil {
 			lg.Traceln("attempting to forward to a closed Accumulator")
 		}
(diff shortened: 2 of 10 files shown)


autholykos:
As it turns out, the race condition has been triggered primarily because the test does not reproduce the whole extent of the reality in production. 

## Marshaled Form Of Messages Is Cached At Reception

Before the messages hit the `Republisher`, they are unmarshaled into a `message.simple` by the `peer.messageRouter` and pushed further. `message.simple` calls `initPayloadBuffer(messageBuffer)` which essentially just caches the marshaled form as it was received by the peer. When `message.Marshal` is triggered, the function first type-checks if it is handling a `message.simple`. If this is the case, the cached buffer gets immediately returned and no actual marshaling happens. 

This means that the republisher is extremely unlikely to ever trigger a race condition, since it unknowngly handles a `message.simple` most (if not all) of the time.

## Marshaling a message.simple cannot trigger a race condition

The `simple.marshaled` as a field is not exposed and can only be accessed during the sequential `Unmarshal` and `Marshal` actions within the `message` package. This means that it cannot be the targe of race conditions between the `Republisher` and the various consensus components, which are both external to the `message` package. 

## Read-only race conditions are possible only within the validation callbacks

The `Republisher` gets injected validation callback functions. These are the only places where concurrent reads can happen (as the validation does not write on the message itself). However, validation functions call-back into the components and therefore read-locks are a responsibility of the components and executed outside of the `Republisher` knowledge.

## Race conditions in the `Republisher` are possible only when the Republisher handles messages other than `message.simple`
 
A message is generally not a `message.simple` instance when it gets generated internally and not received by the peer. However, messages generated internally do not usually get propagated through the `Republisher`, but gossiped directly by the consensus components and therefore they do not really represent a problem. This require double checking though.

So all in all, race-conditions raised by the tests are a false positive

PR #533, opened 2020-05-17
Implement GetUnconfirmedBalance

### LINKED ISSUE #430: Implement `GetUnconfirmedBalance` on the mempool

julesdesmit:
Fixes #430


pkg/core/data/transactions/provider.go
```diff
@@ -226,6 +229,38 @@ func (p *provider) GetBalance(ctx context.Context, vk ViewKey) (uint64, error) {
 	return res.Balance, nil
 }
 
+func (p *provider) CalculateMempoolBalance(ctx context.Context, vkBytes []byte, txs []ContractCall) (uint64, error) {
```
autholykos, line 232: I feel that something so Mempool specific does not belong in the `Provider` interface. I'd rather change the name of `Verifier` interface to `UnconfirmedTxProber` and add `CalculateBalance` there (without `Mempool` which is the component using the prober). This has the advantage that we do not need to inject two different interfaces and testing/mocking would be less verbose

pkg/core/mempool/mempool.go
```diff
@@ -88,7 +89,7 @@ func (m *Mempool) checkTx(tx transactions.ContractCall) error {
 }
 
 // NewMempool instantiates and initializes node mempool
-func NewMempool(ctx context.Context, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier transactions.Verifier, srv *grpc.Server) *Mempool {
+func NewMempool(ctx context.Context, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier transactions.Verifier, provider transactions.Provider, srv *grpc.Server) *Mempool {
```
autholykos, line 92: See previous comment

pkg/core/data/transactions/provider.go
autholykos: I implemented this one myself :)

Issue #473, opened 2020-05-07
Derive deadline and cancellation context within the GRPC methods


autholykos:
Within the provider we simply pass the context. We should consider embedding context handling (with deadline, etc) directly in the provider, instead of relying on the caller

autholykos:
This issue also include changing the `consensus.Component` interface by adding the context to the factories and injecting them in the `Instantiate` method

julesdesmit:
Fixed via #539

PR #539, opened 2020-05-18
Move all deadline cancellations from components to transactions/provider

autholykos:
Also add configuration parameter for the timeout

### DIFF
--- a/pkg/config/samples/default.dusk.toml
+++ b/pkg/config/samples/default.dusk.toml
@@ -96,7 +96,8 @@ keyFile=""
 [rpc.rusk]
 
 # timeout for internal GRPC calls expressed in milliseconds
-timeout= 100
+contractTimeout=6000
+defaultTimeout=200
 
 # GraphQL API service
 [gql]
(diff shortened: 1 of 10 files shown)


autholykos:
Fixes #473

pkg/config/samples/default.dusk.toml
```diff
@@ -92,6 +93,11 @@ certFile=""
 # server TLS key file
 keyFile=""
 
+[rpc.rusk]
+
+# timeout for internal GRPC calls expressed in milliseconds
+timeout= 100
```
julesdesmit, line 99: 100 ms may be much too short in some cases - think for instance the creation of a PLONK proof (known to be several seconds at the moment). I think between 5 and 10 seconds would be more sensible, though it would be even better if we can adjust this on a case-by-case basis, since this lengthy timeout is only useful in a limited amount of cases.

autholykos: ah right, I had in mind the full node (where timeouts should really be kept to a minimum). I will create a `contractTimeout` and a `defaultTimeout`. The former should be enough for transaction creations (which include PLONK), and the default should cover the rest. I don't think we need more granularity than that and we can always add parameters later

julesdesmit: Sounds good!

PR #560, opened 2020-06-12
Enable authenticated session for GRPC 

### LINKED ISSUE #509: Implement some type of session management functionality for the wallet
For security reasons, it would be wise to have session-based availability of the wallet keys. For instance, we can add a gRPC method that unloads the keys and essentially locks up the wallet again, which can be called by front-end apps in order to secure funds after a timeout. 

Another approach could be to not keep the keys inside of a wallet struct, but instead requiring a password every time a user wishes to do a key-based operation, which then decrypts the keys, performs the operation, and then drops them after they go out of scope.

autholykos:
Also includes a simple golang client to be used in the test-harness

Fixes #509 #560 #563


Makefile
```diff
@@ -16,7 +16,7 @@ test-harness-ci: build
 	MOCK_ADDRESS=127.0.0.1:8080 NETWORK_SIZE=3 DUSK_BLOCKCHAIN=${PWD}/bin/dusk DUSK_UTILS=${PWD}/bin/utils DUSK_SEEDER=${PWD}/bin/voucher DUSK_WALLET_PASS="password" make test-harness
 test-harness-alive: stop build
 	MOCK_ADDRESS=127.0.0.1:9191 NETWORK_SIZE=3 DUSK_BLOCKCHAIN=${PWD}/bin/dusk DUSK_UTILS=${PWD}/bin/utils DUSK_SEEDER=${PWD}/bin/voucher DUSK_WALLET_PASS="password" \
-	go test -v --count=1 --test.timeout=0 ./harness/tests/ -args -enable -keepalive
+	go test -v --count=1 --test.timeout=0 ./harness/tests/ -args -enable
```
julesdesmit, line 19: Maybe this was a consequence of merging with `rusk-ipc`, if not, why is the `-keepalive` flag removed?

cmd/dusk/server.go
```diff
@@ -101,11 +101,12 @@ func Setup() *Server {
 	rpcBus := rpcbus.New()
 
 	// Instantiate gRPC client
-	client := rpc.InitRPCClients(ctx, cfg.Get().RPC.Rusk.Address)
+	// TODO: get address from config
```
julesdesmit, line 104: This TODO seems fixed.

harness/engine/profiles.go
```diff
@@ -41,7 +42,7 @@ func Profile1(index int, node *DuskNode, walletPath string) {
 	viper.Set("wallet.store", node.Dir+"/walletDB/")
 	viper.Set("wallet.file", walletPath+"/wallet-"+node.Id+".dat")
 	viper.Set("network.seeder.addresses", []string{"127.0.0.1:8081"})
-	viper.Set("network.port", strconv.Itoa(7100+index))
+	viper.Set("network.port", strconv.Itoa(7600+index))
```
devtomnl, line 45: is this required to change the ports ?

harness/engine/utils.go
```diff
@@ -81,16 +81,14 @@ func (n *Network) SendQuery(nodeIndex uint, query string, result interface{}) er
...
 func (n *Network) LoadWalletCmd(ind uint, password string) (string, error) {
-	addr := "unix://" + n.Nodes[ind].Cfg.RPC.Address
+	// session has been setup already in the TestMain, so here the client
+	// should be returning the permanent connection
+	conn, err := n.Nodes[ind].GRPCClient.GetSessionConn(grpc.WithInsecure(), grpc.WithBlock())
 
-	// Set up a connection to the server.
-	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithBlock())
 	if err != nil {
+		fmt.Println(n.Nodes[ind])
```
devtomnl, line 89: good place for a logger ?

harness/engine/profiles.go
autholykos: not really, no

Makefile
autholykos: it was a mere oversight. Reintroduced

devtomnl:
Unit test is failing with:
```
--- FAIL: TestCreateDropSession (0.00s)

    session_test.go:12: 

        	Error Trace:	session_test.go:12

        	Error:      	Received unexpected error:

        	            	rpc error: code = Unimplemented desc = unknown service node.Auth

        	Test:       	TestCreateDropSession

FAIL
```

PR #570, opened 2020-06-30
Add filebeat + elk support for devnet #569

devtomnl:


devtomnl:
fixed lint issues:
```
pkg/gql/query/transactions.go:5: File is not `goimports`-ed (goimports)
        "encoding/hex"
pkg/core/mempool/mempool.go:581:91: `handleRequest` - `name` is unused (unparam)
func handleRequest(r rpcbus.Request, handler func(r rpcbus.Request) (interface{}, error), name string) {
                                                                                          ^
pkg/core/chain/chain.go:462:5: shadow: declaration of "err" shadows declaration at line 443 (govet)
        if err := c.finalizeIntermediateBlock(c.ctx, cm.Certificate); err != nil {

```

pkg/util/legacy/conversions.go
```diff
@@ -396,6 +396,8 @@ func ruskInputsToInputs(inputs []*rusk.TransactionInput) (transactions.Inputs, e
 	sInputs := make(transactions.Inputs, len(inputs))
 
 	for i, input := range inputs {
+		sInputs[i] = new(transactions.Input)
```
devtomnl, line 399: #597

pkg/util/nativeutils/eventbus/publisher.go
```diff
@@ -17,18 +16,22 @@ type Publisher interface {
 // Publishing is a fire and forget. If there is no listener for a topic, the
 // messages are lost
 func (bus *EventBus) Publish(topic topics.Topic, m message.Message) {
-	logEB.WithFields(logrus.Fields{
-		"topic":    topic,
-		"category": m.Category(),
-	}).Traceln("publishing on the eventbus")
+	//logEB.WithFields(logrus.Fields{
```
devtomnl, line 19: removed the publishing on the eventbus log, since its really just all the time logged and the value of it is questionable, suggestions ?

pkg/p2p/peer/handshake.go
```diff
@@ -12,20 +12,20 @@ import (
 )
 
 // Handshake with another peer.
-func (p *Writer) Handshake() error {
-	if err := p.writeLocalMsgVersion(p.gossip); err != nil {
+func (w *Writer) Handshake() error {
```
devtomnl, line 15: this was renamed to be consistent with other `(w *Writer)`

pkg/gql/query/transactions.go
```diff
@@ -92,27 +91,27 @@ func newQueryTx(tx core.ContractCall, blockHash []byte) (queryTx, error) {
 // IsNil will check for nil in a output
 func IsNil(output *core.TransactionOutput) bool {
 	if output.Pk == nil {
-		log.Warn("invalid output, Pk field is nil")
+		//log.Warn("invalid output, Pk field is nil")
```
devtomnl, line 94: this logs where commented out as it was just polluting the logs with no clear added value. suggestions ?

autholykos: maybe reduce severity to `traces`?

pkg/util/nativeutils/eventbus/publisher.go
autholykos: Fine to remove

pkg/p2p/peer/handshake.go
autholykos: OK

pkg/gql/query/transactions.go
devtomnl: it does not really add anything, this logs really pollute everything when running the metrics collection.

Issue #573, opened 2020-07-03
Implement REST api for healthcheck


devtomnl:

autholykos:
This includes providing information about: 
- amount of active bidders for a given round
- amount and ID of provisioners
- current step the node's consensus algorithm is at (this information is handled by the Coordinator)
- status of the `EventQueue`

autholykos:
Closed via #579

PR #579, opened 2020-07-14
Add Healthcheck API endpoints #573

devtomnl:


autholykos (approved):

pkg/core/chain/chain.go
```diff
@@ -359,11 +359,13 @@ func (c *Chain) AcceptBlock(ctx context.Context, blk block.Block) error {
...
-		if err != nil {
-			log.Warn("Could not store provisioners on memoryDB")
-		}
+		go func() {
+			store := capi.GetBuntStoreInstance()
+			err := store.StoreProvisioners(c.p, blk.Header.Height)
+			if err != nil {
+				log.Warn("Could not store provisioners on memoryDB")
+			}
+		}()
```
autholykos, line 368: I assume the error is limited to the monitoring functions and does not influence the chain logic. If this i not the case (which I don't think), it may be better to use an `errgroup` or an error channel to report back the error, in case `store.StoreProvisioners` would fail

pkg/core/consensus/capi/db.go
```diff
@@ -18,6 +19,7 @@ var (
 	RoundInfoPrefix    = "roundinfo"
 	EventQueuePrefix   = "eventqueue"
 	buntStoreInstance  *BuntStore
+	dbMu               sync.RWMutex
```
autholykos, line 22: The DB operations appear not to be threadsafe? This is rather weird

autholykos (approved):
LGTM. I left a couple of comments about the error handling for the store operations. If my understanding is correct, the PR can be merged

pkg/core/chain/chain.go
autholykos: Confirmed that the `store` is only for the monitoring APIs, therefore the error can be logged in a fire and forget manner

pkg/core/consensus/capi/db.go
autholykos: Confirmed that the mutex can be removed

Issue #582, opened 2020-07-18
An unsynchronized delete and write of bidderList


ggoranov:
`(c *Chain) sendRoundUpdate() ` publishes `topics.RoundUpdate` with payload of bidlist as a reference.

Collectors of `topics.RoundUpdate` and Chain component are concurrently reading/writing `bidlist` slice in a unsafe manner.

```

==> /tmp/localnet-216084718/node-9000/dusk_node_stderr <==
==================
WARNING: DATA RACE
Read at 0x00c000299488 by goroutine 27:
  github.com/dusk-network/dusk-blockchain/pkg/core/chain.(*Chain).removeExpiredBids()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/chain/chain.go:593 +0xbf
  github.com/dusk-network/dusk-blockchain/pkg/core/chain.(*Chain).AcceptBlock()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/chain/chain.go:340 +0x1110
  github.com/dusk-network/dusk-blockchain/pkg/core/chain.(*Chain).handleCertificateMessage()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/chain/chain.go:642 +0x341
  github.com/dusk-network/dusk-blockchain/pkg/core/chain.(*Chain).Listen()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/chain/chain.go:200 +0x65f

Previous write at 0x00c000299488 by goroutine 118:
  github.com/dusk-network/dusk-blockchain/pkg/core/consensus/user.BidList.Subset.func1()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/consensus/user/bidlist.go:65 +0x1a7
  math/rand.(*Rand).Shuffle()
      /usr/local/go/src/math/rand/rand.go:253 +0xc4
  github.com/dusk-network/dusk-blockchain/pkg/core/consensus/user.BidList.Subset()
      /usr/local/go/src/math/rand/rand.go:346 +0x99
  github.com/dusk-network/dusk-blockchain/pkg/core/consensus/generation/score.createBidListSubset()
      /home/tech/repo/dusknetwork/temp/dusk-blockchain/pkg/core/consensus/generation/score/generator.go:162 +0x62
  github.com/dusk-network/dusk-blockchain/pkg/core/consensus/generation/score.(*Generator).Prove()
[...]

PR #540, opened 2020-05-18
Implement architecture for safe message passing

autholykos:
Fixes #292

PR #583, opened 2020-07-18
Consensus/fix race

ggoranov:
Fixes https://github.com/dusk-network/dusk-blockchain/issues/582


autholykos:
This is likely already fixed in `rusk-ipc` due to the changes introduced to #540 which introduce implicit deep copy on `EventBus` messages. Still there is value in the test-harness improvements

PR #591, opened 2020-07-24
add option to configure consensustimeout

devtomnl:
add option to configure consensustimeout as oposite of having hardcoded 5 seconds to enable testing it easily

### DIFF
--- a/pkg/config/samples/default.dusk.toml
+++ b/pkg/config/samples/default.dusk.toml
@@ -153,6 +153,8 @@ accumulatorWorkers = 4
 defaultlocktime = 250000
 # default amount, in whole units of DUSK, to send for consensus transactions.
 defaultamount = 5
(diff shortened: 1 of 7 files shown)


pkg/config/samples/default.dusk.toml
```diff
@@ -153,6 +153,8 @@ accumulatorWorkers = 4
 defaultlocktime = 250000
 # default amount, in whole units of DUSK, to send for consensus transactions.
 defaultamount = 5
+# the timeout for consensus step timers
+consensustimeout = 1
```
ggoranov, line 157: Default `ConsensusTimeout` used to be 5 sec (as per `config.ConsensusTimeOut`).  Why do you change it to 1 sec?

autholykos: Good catch, will fix it in #579

PR #599, opened 2020-07-29
Fix issue 592

ggoranov:
How to test: `go test github.com/dusk-network/dusk-blockchain/pkg/core/chain/... -v -race -run TestConcurrentBlock`


harness/tests/localnet_test.go
```diff
@@ -107,7 +107,14 @@ func TestSendBidTransaction(t *testing.T) {
 	t.Log("Send request to node 0 to generate and process a Bid transaction")
 	txidBytes, err := localNet.SendBidCmd(0, 10, 10)
 	if err != nil {
-		t.Fatal(err.Error())
+		t.Fatal(err)
+	}
+
+	for i := 1; i < localNet.Size(); i++ {
```
devtomnl, line 113: so we are now going to always test with sending bids to all participants by default ?

pkg/core/chain/chain.go
```diff
@@ -200,17 +202,26 @@ func New(ctx context.Context, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus
...
 	return chain, nil
 }
 
 // Listen to the collectors
 func (c *Chain) Listen() {
 	for {
 		select {
+		case m := <-c.blockChan:
+			if err := c.onAcceptBlock(m); err != nil {
+				log.WithError(err).Warnf("Handling %s failed", topics.Block.String())
```
devtomnl, line 219: could you please log it without Warnf and split the fields? that helps with the monitoring.

pkg/core/chain/chain.go
```diff
@@ -200,17 +202,26 @@ func New(ctx context.Context, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus
...
+			}
+		case m := <-c.initializationChan:
+			if err := c.onInitialization(m); err != nil {
+				log.WithError(err).Warnf("Handling %s failed", topics.Initialization.String())
```
devtomnl, line 223: same here

pkg/util/nativeutils/eventbus/listenermap.go
```diff
@@ -36,8 +36,10 @@ func (h *listenerMap) Store(key topics.Topic, value Listener) uint32 {
 func (h *listenerMap) Load(key topics.Topic) []idListener {
 	h.lock.RLock()
 	listeners := h.listeners[key]
+	dup := make([]idListener, len(listeners))
+	copy(dup, listeners)
 	h.lock.RUnlock()
```
devtomnl, line 41: could we add the unlock with defer on the begin of the func ? just for consistency.

harness/tests/localnet_test.go
ggoranov: Yes, that's the idea. Such situation is closer to what you can expect in mainnet (multiple bidders network)

pkg/util/nativeutils/eventbus/listenermap.go
ggoranov: No, defer is not needed here.

pkg/core/chain/chain.go
ggoranov: No, this still can be easily parsed by monitoring.

devtomnl:
LGTM, all tests are passing.

PR #610, opened 2020-08-04
Update the signature of the method Publish in the Publisher interface to also return error so issues can also be properly logged #609

devtomnl:
Update method signature of (m *multiListener) Forward to include errorList []error

Update Publisher Publish method to return []error and update all places where it is used with proper err log


julesdesmit:
Once the CI failures are addressed, this is fine with me to be merged.

Issue #626, opened 2020-08-12
Selection should store/update the Candidate associated with the highest seen score for the current Step


autholykos:
1. BlockGenerator component should propagate `topics.Score` together with both `Candidate Block` and `Certificate` in a single wire message
	- BlockGenerator should NOT propagate `topics.Candidate` anymore
2. On handling `topics.Score` Selection should 
	- if score is bigger than current bestScore
		- verify score, if true
		- sanity check the candidate block, if true
		- swap bestScore
		- swap bestCandidate(Block and Certificate)
3. On Selection timeout triggered 
         - Publish to network a `topics.Candidate` msg with bestCandidate (block + certificate)
         - Publish (internally) `topics.BestEvent` so reduction firststep can be started with bestCandidate blockhash

PR #633, opened 2020-08-19
Selection should store/update the Candidate associated with the highest seen score for the current Step #626

devtomnl:
closes: #626


ggoranov:
Identical with `fix/issue_626`

PR #635, opened 2020-08-21
Fix/issue 626

ggoranov:


pkg/core/consensus/selection/selector.go
```diff
@@ -128,6 +128,20 @@ func (s *Selector) CollectScoreEvent(packet consensus.InternalPacket) error {
...
+		return nil
+	}
+
+	// Publish internally topics.Candidate with bestEvent(highest score candidate block)
+	msg := message.New(topics.Candidate, score.Candidate)
+	err := s.signer.SendInternally(topics.Candidate, msg, s.ID())
+	if err != nil {
+		lg.WithError(err).Error("CollectScoreEvent, failed to SendInternally the Candidate")
+	}
+
```
autholykos, line 144: Would it make sense to use the RPCBus to avoid excessive asynchronism?

autholykos (approved):

PR #652, opened 2020-09-07
Fix/issue 642

ggoranov:


autholykos (approved):
LGTM. I would prefer the `parameter` preparation to be extrapolated in a dedicated function to prevent users "forgetting" to set the boolean flag (which is used for execution control rather than a real parameter), but this can be done in a separate, minor, issue

Issue #654, opened 2020-09-08
Upgrade lint to latest and fix / create tasks for new reported lint issues


devtomnl:

devtomnl:
```
pkg/core/consensus/header/header.go:233:10: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        step := rand.Intn(8)
                ^
pkg/p2p/kadcast/utils.go:322:9: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        ind := rand.Intn(maxVal)
               ^
pkg/core/data/transactions/fixtures.go:166:10: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
                idx := mrand.Intn(amount)
                       ^
pkg/core/data/transactions/fixtures.go:650:7: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        t := mrand.Intn(8)
             ^
pkg/util/nativeutils/eventbus/listener.go:206:13: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
                id:       rand.Uint32(),
                          ^
pkg/util/nativeutils/eventbus/listenermap.go:28:8: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        id := rand.Uint32()
              ^
pkg/core/consensus/component.go:102:8: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        id := rand.Uint32()
              ^
pkg/core/consensus/component.go:152:8: G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
        id := rand.Uint32()
              ^

```

PR #655, opened 2020-09-08
 Upgrade lint to latest and fix / create tasks for new reported lint issues #654 

devtomnl:


julesdesmit (commented):
The test-harness run should be fixed, and the conflicts resolved before merging.

Issue #603, opened 2020-07-31
Update CI and build process to use Go 1.14


devtomnl:

devtomnl:
Update locks to use defer to be done in a separated ticket, if ever.

PR #666, opened 2020-09-11
Update locks to use defer #603

devtomnl:
* add defer to all lock operations, wherever possible. 
* few places I did not change and added TODO for further discussion on what to do


devtomnl:
This was placed in the icebox for now since its a cosmetic change.

PR #667, opened 2020-09-14
Fix/issue 659

ggoranov:


pkg/core/chain/chain.go
```diff
@@ -355,9 +345,19 @@ func (c *Chain) AcceptBlock(ctx context.Context, blk block.Block) error {
...
+	// 3. Check the certificate
+	// This check should avoid a possible race condition between accepting two blocks
+	// at the same height, as the probability of the committee creating two valid certificates
+	// for the same round is negligible.
+	l.WithField("provisioners_len", c.p.Set.Len()).Trace("verifying block certificate")
+	if err := verifiers.CheckBlockCertificate(*c.p, blk); err != nil {
+		l.WithError(err).Error("certificate verification failed")
+		return err
+	}
+
```
autholykos, line 360: The order of execution seems wrong. If the certificate is invalid, we should not call `ExecuteStateTransition` function, since that function will mutate the smart contract storages with no possibility to rollback

autholykos (changes requested):

pkg/core/chain/chain.go
autholykos: The wrong execution order is confirmed. In order to solve the race condition, a better solution has been found and is blocked by #672

ggoranov:
This PR is closed as another branch (based on latest rusk-ipc) has been created to address issue 659

PR #687, opened 2020-09-22
p2p: readWriteTimeout and keepAliveTime should be configurable #686

devtomnl:
update peer.go to have its main timeout variables configurable by adding timeoutreadwrite and timeoutkeepalivetime
update action.go profiles.go dusk.toml groups.go to reflect the change
update handshake.go receiver name from p to c in order to have the same receiver name as in other places and make it uniform
update handshake_test.go and peer_in_test.go by adding viper params to it
update devnet.sh adding new configs to it

pending fix: #692


pkg/config/samples/default.dusk.toml
```diff
@@ -22,6 +22,8 @@ timeoutsendstaketx = 5
 timeoutgetmempooltxs = 3
 timeoutgetroundresults = 5
 timeoutbrokergetcandidate = 2
+timeoutreadwrite = 60
+timeoutkeepalivetime = 30
```
ggoranov, line 26: It should be good to clarify here as a comment that `timeoutkeepalivetime`  must be always smaller than `timeoutreadwrite`, otherwise the node will disconnect due to `read timeout error`

devtomnl: sure

julesdesmit (approved):
LGTM

Issue #688, opened 2020-09-23
Replace BoltDB queries with Storm


devtomnl:
https://github.com/asdine/storm

```
Storm is a simple and powerful toolkit for BoltDB. Basically, Storm provides indexes, a wide range of methods to store and fetch data, an advanced query system, and much more.
```

devtomnl:
motivation:
To maintain the existing code is way too much work, so using Storm things are way more simplified, specially for finding operations.

julesdesmit:
Closed via #689

Issue #680, opened 2020-09-21
p2p/Peer should expose a endpoint to be queried that returns count of connected peers and detailed list of peers


devtomnl:

julesdesmit:
Closed via #689

PR #689, opened 2020-09-23
p2p/Peer should expose a endpoint to be queried that returns count of connected peers and detailed list of peers #680

devtomnl:
Replace BoltDB queries with Storm #688


julesdesmit (commented):
LGTM, but make sure to fix the lint first.

devtomnl:
> LGTM, but make sure to fix the lint first.

sure

PR #726, opened 2020-10-27
Consensus/testing

ggoranov:


pkg/core/consensus/reduction/secondstep/step.go
```diff
@@ -53,10 +60,13 @@ func (p *Phase) Fn(re consensus.InternalPacket) consensus.PhaseFn {
...
 // of votes, or we experience an unrecoverable error
 func (p *Phase) Run(ctx context.Context, queue *consensus.Queue, evChan chan message.Message, r consensus.RoundUpdate, step uint8) consensus.PhaseFn {
-	lg.
-		WithField("round", r.Round).
-		WithField("step", step).
-		Trace("starting secondstep reduction")
+	tlog := getLog(r.Round, step)
+	tlog.Traceln("starting second reduction step")
+
+	defer func() {
```
ggoranov, line 66: Maybe if we introduce Phase name field we can move begin/end trace logs in consensus.spin for all phases

autholykos (approved):
LGTM

PR #741, opened 2020-11-04
Change the PhaseFn from function to interface so to ease logging in the consensus loop

### LINKED ISSUE #740: centralize logging of step execution in the consensus loop
It would be good to let the consensus loop log the name of phase it is executing. This is not currently possible since the consensus phase function returns a `PhaseFn` which is still a function. 

There are two alternatives to allow this. The first is to change the `PhaseFn` to return the next phase function and the name. This is quite ugly since it mixes the logic of the phase execution with that of logging). The second could be to return `Phase` from `PhaseFn` and refactor the recursion around the `Phase` struct rather than `PhaseFn` function

autholykos:
Fixes #740


ggoranov (approved):
LGTM

Issue #743, opened 2020-11-04
Add flags to Voucher to bind to IP and Port


devtomnl:

julesdesmit:
These flags already seem to have been implemented (probably merged without linking the issue). So, I'm closing this.

PR #745, opened 2020-11-04
Add flags to Voucher to bind to IP and Port #743

devtomnl:
* Add flags for Port and Hostname
* fix log file for rusk mock


julesdesmit (approved):
LGTM

PR #746, opened 2020-11-05
Chain/synchronizer refactor

julesdesmit:


pkg/p2p/peer/peer.go
```diff
@@ -316,17 +273,19 @@ func (p *Reader) readLoop() {
...
-		// TODO: error here should be checked in order to decrease reputation
-		// or blacklist spammers
-		startTime := time.Now().UnixNano()
-		if err := p.router.Collect(message); err != nil {
-			l.WithError(err).Error("message routing")
-		}
+		go func() {
+			// TODO: error here should be checked in order to decrease reputation
+			// or blacklist spammers
+			startTime := time.Now().UnixNano()
```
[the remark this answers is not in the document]
ggoranov: It's not boilerplate to test such scenario so I raise it as a question before ensuring its a real thing.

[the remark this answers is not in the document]
julesdesmit: While it is true that messages will be handled in a random order due to this approach, at first thought, I don't think this will cause any issues. `Reader`s do not expect messages to come in order, and they shouldn't either. Once a message is received, its processing is handled in a separate goroutine to all others, which means that if both a `Ping` and `GetData` are sent at the same time, they will always end up in the right place.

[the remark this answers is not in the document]
julesdesmit: It may however be good to test this, which I will spend some effort on.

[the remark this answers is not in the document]
ggoranov: I'd advice to test this with enough attention. Go scheduler is non-deterministic.

pkg/core/chain/chain.go
```diff
@@ -176,115 +152,138 @@ func New(ctx context.Context, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus
...
+		c.highestSeen = blk.Header.Height
 	}
 
 	// If we are more than one block behind, stop the consensus
-	lg.Debug("topics.StopConsensus")
+	log.Debug("topics.StopConsensus")
 	// FIXME: this call should be blocking
-	errList := c.eventBus.Publish(topics.StopConsensus, message.New(topics.StopConsensus, message.EMPTY))
-	diagnostics.LogPublishErrors("chain/chain.go, topics.StopConsensus", errList)
+	if c.cancel != nil {
```
ggoranov, line 205: Is it true that any node who sends a forged block with higher block height is capable of cancel/disabling consensus loop for awhile? 

Without a sanity or certificate check for the wire (block) message, a malicious  node might be capable to suspend consensus for majority of the network for the benefit of his/her own provisioners/generators.  

NB I raised this on a few stand-ups but not sure if it is ignored/forgotten.

julesdesmit: That's right, but I don't think this is an issue with the architecture per se. With the old `Synchronizer` approach, the same problem occurs. Perhaps this is something we should extract into another issue?

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
julesdesmit: Right, however it is also true that the `MessageProcessor` will never allow messages to be crossed up, as explained in the chat thread. As mentioned, it may be an issue for messages that are only of the same topic, but this is more component-specific and thus I think the peer layer shouldn't be concerned with ordering messages. An example of that is the `Chain`, which solved it with the `sequencer` component.

cmd/dusk/server.go
```diff
@@ -100,7 +103,28 @@ func Setup() *Server {
...
+	// Register peer services
+	processor.Register(topics.Ping, responding.ProcessPing)
+	dataBroker := responding.NewDataBroker(db, rpcBus)
+	processor.Register(topics.GetData, dataBroker.SendItems)
+	dataRequestor := responding.NewDataRequestor(db, rpcBus, eventBus)
+	processor.Register(topics.Inv, dataRequestor.RequestMissingItems)
+	bhb := responding.NewBlockHashBroker(db)
+	processor.Register(topics.GetBlocks, bhb.AdvertiseMissingBlocks)
+	cb := responding.NewCandidateBroker(db)
+	processor.Register(topics.GetCandidate, cb.ProvideCandidate)
```
ggoranov, line 117: Is any component gossiping `topics.Candidate` anywhere?  As per protocol, we fetch the candidate from network, if it cannot be retrieved from local state.

pkg/core/database/heavy/transactions.go
```diff
@@ -591,6 +593,45 @@ func (t transaction) FetchBlockHeightSince(sinceUnixTime int64, offset uint64) (
...
+	if err := message.MarshalCandidate(buf, cm); err != nil {
+		return err
+	}
+
+	key := append(CandidatePrefix, cm.Block.Header.Hash...)
+	t.put(key, buf.Bytes())
+	return nil
+}
+
+func (t transaction) FetchCandidateMessage(hash []byte) (*message.Candidate, error) {
```
ggoranov, line 608: Not sure why we're going back again to the idea of storing candidate blocks per round in the blockchain database (persistently) instead of keeping them as a cache This has its drawbacks.

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
ggoranov: This problematic scenario is expected to be reproducible with `GetCandidate` request-response messaging.

cmd/dusk/server.go
julesdesmit: Ah, I left the code commented. I can add it back to the candidate broker.

pkg/core/chain/chain.go
ggoranov: It's an architecture thing but we always can postpone it.

pkg/core/database/heavy/transactions.go
julesdesmit: Mostly for consistency with the rest of the design. Using the DB facilitates simplicity, and reduces concurrency which wasn't entirely necessary, considering that the DB is easy to access for any part of the node. As the DB is ready for concurrent access, there should not be any danger in using it for this purpose.

However, it is only one part of the larger refactor. If there is enough reason to reintroduce concurrency for fetching candidates, we can revert to it if needed.

pkg/p2p/peer/processor.go
```diff
@@ -0,0 +1,92 @@
...
+
+// ProcessorFunc defines an interface for callbacks which can be registered
+// to the MessageProcessor, in order to process messages from the network.
+type ProcessorFunc func(message.Message) ([]*bytes.Buffer, error)
+
+// MessageProcessor is connected to all of the processing units that are tied to the peer.
+// It sends an incoming message in the right direction, according to its topic.
+type MessageProcessor struct {
+	dupeMap    *dupemap.DupeMap
+	processors map[topics.Topic]ProcessorFunc
```
ggoranov, line 22: This is another form of `CallbackListener` from EventBus.  So now we have P2P sending calls to the upper layer directly (via MessageRouter callbacks) but all other layers (chain/mempool/ ...) are sending msg to the P2P via EventBus.Publish. No consistency here.

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
julesdesmit: Could you elaborate?

pkg/core/database/heavy/transactions.go
ggoranov: You have many options here different from reusing the blockchain db instance. Please do not overload the blockchain db with intermediate stuff like candidates. 

If your idea was simply to reduce concurrency (no broker loop), let's keep the approach of in-memory candidates store and provide again direct access (in non-rpcbus manner)

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
ggoranov: Once you restore the part for `GetCandidate from the network`, please provide a test ensuring that GetCandidate request is not responded with non-candidate payload due to the concern stated above.

pkg/p2p/peer/processor.go
julesdesmit: Not sure if there needs to be consistency there, keep in mind that receiving a message from a single peer is a bit different to gossiping a message to the entire network. The event bus is a simple way for components to basically just put items on a set of ring buffers. 

IMO, in this scenario, the event bus makes more sense due to the multi-cast nature of the message, whereas the processing of a single message is fine to handle directly through a callback from the message processor. We can use CallbackListeners if the consistency is really desired, but to me, it seems like an unnecessary layer of indirection, since we are always speaking to a single component.

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
julesdesmit: Like I said, messages that are not intended for the recipient will not end up there. When a candidate is requested, the broker will listen on `topics.Candidate` for incoming messages, so there is no conceivable way a `Ping` would end up in the broker, for instance. Since the `peer` is also not aware of ongoing requests and responses, there should be no way of mixing up messages. I can provide the test if you prefer, but I don't see how this would be possible.

pkg/p2p/peer/processor.go
julesdesmit: In contrast, the direct responses that need to be made will result as a return argument from the callback that was called - which alleviates the need for a more complex design, i.e. the router, which needed to be instantiated for every peer, and the router itself needed to instantiate lots of different structs that were specifically used to respond directly to peers, with different topics being routed needing one of two dissemination techniques.

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
ggoranov: If you want to say that it is safe to correlate request and response only on the topic type, this is pretty wrong assumption. I suggested on the stand-up to introduce request_id (like stream identifier in HTTP/2). if you want to support concurrency in P2P message handling that's the idiomatic approach. (just not sure if this is the idea of this PR)

pkg/p2p/peer/peer.go
```diff
@@ -316,17 +273,19 @@ func (p *Reader) readLoop() {
 			return
 		}
 
-		// TODO: error here should be checked in order to decrease reputation
-		// or blacklist spammers
-		startTime := time.Now().UnixNano()
-		if err := p.router.Collect(message); err != nil {
-			l.WithError(err).Error("message routing")
-		}
+		go func() {
```
ggoranov, line 276: Spawning a goroutine to handle a request is pretty idiomatic approach. But again if the idea of this PR is to reduce concurrency not sure how this is achieved if you're introducing hundreds of new goroutines (consider 9 peer connections processing at least 10 wire messages at a time )

ggoranov: In addition, not sure we need to gain more technical debt in P2P layer as we tend to migrate to Kadcast.

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
autholykos: While I also do not see an immediate correlation problem, I also do agree that adding an ID to correlate requests and response is pretty inexpensive, would help solving potential problems and as an approach could be expanded in the future. 

However, this is not part of this PR. @ggoranov please open an issue for this and detail how an idiomatic approach would look like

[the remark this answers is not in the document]
julesdesmit: I'm not just assuming it though; this is testable through the keepalive test harness, where request-response communication happens when transactions are created and blocks are produced. The handling of these kinds of messages, on a high level, has not really changed with this refactor, so I don't know why this wasn't brought up earlier. Where a `topics.Block` was originally passed through a channel by the `Synchronizer`, it is now handled through a callback in a goroutine. In both scenarios, the block processing is happening concurrently to the peer reading out new messages. 

If the peer reads out a message with `topics.GetCandidate`, it is processed by a callback which is registered to that topic, not the callback for `topics.Block` or anything else. The response topic is entirely dictated by the callback that is registered, so it would be a programming error if the callback responded on the wrong topic, and not an ephemeral one. If the message does not match the topic, the unmarshalling of the message should fail, and thus not even be processed further. So, I'm not just saying it is safe on a wild assumption.

However, as I seem unable to convince you of its safety, I will add a test once the candidate request procedure is added back.

pkg/core/chain/chain.go
autholykos: Agreed on postponing, but the issue must be recorded. @ggoranov would you mind creating an issue for this?

ggoranov:
To recap my comments, the PR is a good attempt  but it introduces new technical debt. It would be good also to provide some profiling data  proving the new approach reduces the node concurrency. A simple check would be to count number of background long-lived goroutines  before and after this PR.

pkg/p2p/peer/peer.go
julesdesmit: To be more clear - the goal is to reduce concurreny where it is not necessary, so as to avoid potential race conditions, and simplify code. If concurrency can be used to its advantage, there is definitely room for it, which is why I use it here. For this same reason, I did not get rid of inter-process communication performed between, for instance, the mempool and the blockgenerator. Neither did I get rid of the channels used in the consensus for event processing, since the pattern actually works really well there and I don't think it would benefit from a different design. You've said it yourself, spawning goroutines to handle requests is an idiomatic approach, and it was preferred this way when it was mentioned on the stand-up a few days ago, to avoid bottlenecking. It does increase the number of goroutines, but the processing pipeline for many of the messages have become synchronous in isolation, the aim of which, again, was to avoid potential data races.

pkg/core/database/heavy/transactions.go
autholykos: I was the one suggesting the use of the DB. In general, DB are designed to withstand thousands of transactions per second without breaking a sweat, so I don't see us hitting an overloading scenario by adding tens of transaction (i.e. storing 6-12 candidates at every selection run). This will also help in eventual post-mortem. 

Caching mechanisms will be reintroduced later on with a dedicated PR once we are convinced that all stability issues are solved

pkg/p2p/peer/processor.go
autholykos: The `EventBus` is for fire and forget and support N:M message passing. In the case of `MessageProcessor`, if I understand the design, the `ProcessorFunc` is expected to return a value and therefore it is no asynchrony and the communication is 1:1. This would be modeled through the `RPCBus` rather then plumbing `CallbackListener`, however I do not see the need to complicate the matter by adding the `RPCBus`

pkg/p2p/peer/peer.go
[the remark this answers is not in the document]
autholykos: I also trust no-one without a test ;)

pkg/core/consensus/testing/consensus_test.go
```diff
@@ -26,7 +27,7 @@ func mockGenesis() (block.Block, block.Certificate) {
 // TestConsensus passing means the consensus phases are properly assembled
 func TestConsensus(t *stdtesting.T) {
```
ggoranov, line 28: Is there a reason to disable the entire `TestConsensus` including the sandbox environment for testing consensus?

julesdesmit: Mostly because it didn't work with the design at one point in the refactor. I also figured, that maybe the test-harness served the same purpose, and this was created to enable quicker integration testing. If it adds extra testing value, I can add it back.

autholykos: I think it would be valuable to have a testbed for consensus that does not rely on the network and focuses solely on the consensus/chain. Corner cases and other scenarios might be simpler to be observed. Will create an issue about it

pkg/core/chain/chain.go
```diff
@@ -411,24 +407,83 @@ func (c *Chain) startConsensus() error {
...
+
+	return cm, err
+}
+
+func (c *Chain) requestCandidate(hash []byte) (message.Candidate, error) {
+	// Make sure we get temporarily notified of incoming messages
+	candidateChan := make(chan message.Message, 10)
+	cChan := eventbus.NewChanListener(candidateChan)
+	id := c.eventBus.Subscribe(topics.Candidate, cChan)
+	defer c.eventBus.Unsubscribe(topics.Candidate, id)
```
autholykos, line 457: This is pretty ugly and signals that `EventBus` subscription and asynchronous notification in general is not the answer. What you might want to consider here is a RPCBus call with a timeout (maybe managed through a context cancelation) to a different component that handles the peer requests to the candidates and is completely transparent to the `Chain`. Such component would deal with the `Candidate` topic. 

Moreover, I am unsure why this logic is in the Chain. I would argue that maybe we should keep the logic related to candidate fetching in one place, rather than scattering it around. The most suitable place might be the consensus, since it must deal with the block candidate verification. To that respect, we could change the `loop.Spin` function to return the certificates and the candidate block. This way, the Chain would not be concerned with that.

autholykos (changes requested):

pkg/p2p/peer/processor.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/peer/dupemap"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message"
+	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/topics"
+	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/eventbus"
+)
+
+// ProcessorFunc defines an interface for callbacks which can be registered
+// to the MessageProcessor, in order to process messages from the network.
+type ProcessorFunc func(message.Message) ([]*bytes.Buffer, error)
```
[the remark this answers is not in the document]
julesdesmit: Yep, definitely could. Will adjust

pkg/core/candidate/publisher.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewBroker returns an initialized Publisher struct.
+func NewBroker(publisher eventbus.Publisher) *Publisher {
+	return &Publisher{
+		publisher: publisher,
+	}
+}
+
+// Process a received Candidate message. Invalid Candidate messages are discarded.
+func (b *Publisher) Process(msg message.Message) ([]*bytes.Buffer, error) {
```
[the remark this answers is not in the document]
julesdesmit: Yes, it's purely just to conform to the interface. However, this function will also change with the refactor of the `GetCandidate` procedure.

pkg/core/candidate/requestor.go
```diff
@@ -0,0 +1,120 @@
...
+				m.RespChan <- rpcbus.NewResponse(nil, err)
+				continue
+			}
+
+			m.RespChan <- rpcbus.NewResponse(cm, nil)
+		case <-ctx.Done():
+			return
+		}
+	}
+}
```
autholykos, line 63: Remove this, pass the `Requestor` directly to the consensus and let the consensus steps call directly `Requestor.RequestCandidate`

autholykos (changes requested):

PR #772, opened 2020-11-19
Remove Certificate from Candidate messages

### LINKED ISSUE #639: Remove block certificates from messages
By removing the intermediate block, there is no longer a need to send certificates since they are deterministic and can be stored locally upon block acceptance

julesdesmit:
Fixes #639


pkg/p2p/wire/message/candidate.go
```diff
@@ -6,24 +6,19 @@ import (
...
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/header"
-	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/key"
-	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/user"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/data/block"
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message/payload"
-	"github.com/dusk-network/dusk-blockchain/pkg/util"
 )
 
 // Candidate is the composition of block and certificates
 type Candidate struct {
```
autholykos, line 14: I guess the Candidate is just *block.Block. No need to add yet another type for the same thing unless there is some compelling reasons I am not seeing

autholykos (changes requested):
From a first glance #639 seems only partially satisfied. The removal of the certificate seems to be implemented correctly, but I cannot see the deterministic certificate generation happening locally.

PR #776, opened 2020-11-20
Replace usage of exitChan for context in Peer

### LINKED ISSUE #761: Consider using a context cancelation in the peer.Writer/Reader (instead of ad-hoc exitChan)

julesdesmit:
This allows either goroutine to cancel the other
on an encountered error. This gives way
to more idiomatic management of these goroutines.

Fixes #761

### DIFF
--- a/pkg/p2p/peer/peer.go
+++ b/pkg/p2p/peer/peer.go
@@ -2,6 +2,7 @@ package peer
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"errors"
 	"io"
@@ -64,7 +65,6 @@ type Reader struct {
 	*Connection
 	processor    *MessageProcessor
 	responseChan chan<- bytes.Buffer
-	exitChan     chan<- struct{} // Way to kill the WriteLoop
 	// TODO: add service flag
 }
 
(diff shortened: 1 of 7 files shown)


pkg/p2p/peer/peer.go
```diff
@@ -180,18 +180,31 @@ func (p *Reader) Accept() error {
 	return nil
 }
 
-// Serve utilizes two different methods for writing to the open connection
-func (w *Writer) Serve(writeQueueChan <-chan bytes.Buffer, exitChan chan struct{}) {
-	defer w.onDisconnect()
+// Create two-way communication with a peer. This function serves to split up
+// the passed context and allowing either goroutine to cancel the other
+// in case of errors.
+func Create(ctx context.Context, reader *Reader, writer *Writer, writeQueueChan <-chan bytes.Buffer) {
```
autholykos, line 186: The context should be passed to the functions doing the work (in this case `reader.ReadLoop` and `writer.Serve`), and not state-wide to a struct instance or a closure. Also, I don't like tightly coupled goroutines and I would advocate for communicating results only to the caller rather than influence each other with a direct cancelation.

Something like the following:

```golang
func Create(ctx context.Context, reader *Reader, writer *Writer, writeQueueChan <-chan bytes.Buffer) {
    ...
    ctx, cancel := context.WithCancel(ctx)
    defer cancel()

    errChan := make(chan error, 1)
    
    go reader.ReadLoop(ctx, writeQueueChan, errChan)
    go writer.WriteLoop(ctx, writeQueueChan, errChan)

    err := <-errChan
    if !err.Is(ErrDisconnect) {
        // log it
    }
}
``` 
if you want to bubble up any error to the server in the `OnAccept` you can do it easily by passing the `errChan` similarly to what you do with `writeQueueChan`. In this case, the `ErrDisconnect` should be known project-wise

pkg/p2p/peer/peer.go
```diff
@@ -350,23 +370,21 @@ func (p *Reader) readLoop() {
 				}
 			}()
 		}
-
 	}
 }
 
-func (p *Reader) keepAliveLoop() (*time.Timer, chan struct{}) {
+func (p *Reader) keepAliveLoop(ctx context.Context) *time.Timer {
```
autholykos, line 376: In case of adoption of the `errChan`, you might wanna pass it to this function. Also, since `timer` is used outside of this function within the `readLoop`, I would rather prefer the latter to create the timer and pass it to `keepAliveLoop` function (which would no longer need to embed a goroutine)

autholykos (changes requested):

pkg/p2p/peer/peer.go
```diff
@@ -180,31 +180,45 @@ func (p *Reader) Accept() error {
...
-	go writer.Serve(writeCtx, cancelReader, writeQueueChan)
+	errChan := make(chan error, 1)
+
+	go reader.ReadLoop(ctx, errChan)
+	go writer.Serve(ctx, writeQueueChan, errChan)
+
+	// Errors are already getting logged with the proper fields attached.
+	// So, we will just wait for one error to come through, after which we can
+	// cancel the context and exit this function.
+	<-errChan
```
autholykos, line 198: The point of `errChan` is to handle errors (among which disconnections). This should be the responsibility of the caller (`Create`) to do. In some cases, we might have some errors that do not grant a disconnection (so a `cancel`). In any case, this is a bit of a nitpick and I will record it in a separate issue

autholykos (approved):
LGTM

PR #789, opened 2020-12-10
Decouple chain consensus

### LINKED ISSUE #788: Incorporate candidate.Requestor into the consensus struct
The Requestor is used solely by the consensus, therefore it does not make much sense to have it as a floating components passed around in the Chain only because it needs to be wired to the `MessageProcessor`. A better alternative would be to encapsulate it into the `loop.Consensus` struct and make `CreateStateMachine` a method of `Consensus` rather than a function. This way the consensus can be added directly to the `MessageProcessor`, delegate to the Requestor and have no loose ends floating around.

This also applies to the `jubjub` `PublicKey` btw

autholykos:
Fixes #788


autholykos:
Duplicate of #793

PR #792, opened 2020-12-11
Sync refactor

### LINKED ISSUE #788: Incorporate candidate.Requestor into the consensus struct

### LINKED ISSUE #790: Add inSync/deSync state machine in the chain.Synchronizer
Following the state machine example of `loop.Phase`, similarly the sync/async state of the Synchronizer is best modeled through a state machine function.

autholykos:
Refactor the synchronizer, chain and consensus in order to achieve decoupling, better readability and overall code clarity

Fixes #788 
Fixes #790


julesdesmit (approved):
LGTM. This is fine to merge for me, once the lint and tests pass (since I'm still looking towards the test-harness myself).

PR #793, opened 2020-12-15
Let caller of peer.Create handle errors

### LINKED ISSUE #782: Peer should handle the error channel in the Create function
See https://github.com/dusk-network/dusk-blockchain/pull/776#discussion_r530253219

julesdesmit:
Fixes #782

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -250,7 +250,12 @@ func (s *Server) OnAccept(conn net.Conn) {
 	logServer.WithField("address", peerReader.Addr()).Debugln("connection established")
 
 	peerWriter := peer.NewWriter(conn, s.gossip, s.eventBus)
-	go peer.Create(context.Background(), peerReader, peerWriter, writeQueueChan)
+	go func() {
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+		err := peer.Create(ctx, peerReader, peerWriter, writeQueueChan)
+		logServer.WithError(err).Warnln("error received from peer")
+	}()
 }
 
 // OnConnection is the callback for writing to the peers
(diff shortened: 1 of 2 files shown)


cmd/dusk/server.go
```diff
@@ -271,7 +273,17 @@ func (s *Server) OnConnection(conn net.Conn, addr string) {
 		log.Panic(err)
 	}
 
-	go peer.Create(context.Background(), peerReader, peerWriter, writeQueueChan)
+	ctx, cancel := context.WithCancel(context.Background())
+	errChan := peer.Create(ctx, peerReader, peerWriter, writeQueueChan)
+	go handlePeerError(errChan, cancel)
```
autholykos, line 278: Returning channels is not very idiomatic in go. IMO, a more readable version would be:

```
func (s *Server) OnAccept(conn net.Conn) {
    ...
    go func() {
        ctx, cancel := context.WithCancel(context.Background())
        defer cancel()
        err := peer.Create(ctx, peerReader, peerWriter, writeQueueChan)
        logServer.WithError(err).Warnln("error received from peer")
    }
    ...
}

func Create(ctx context.Context, reader *Reader, writer *Writer, writeQueueChan <-chan bytes.Buffer) chan error {
        errChan := make(chan error, 1)
        defer close(errChan)

	go reader.ReadLoop(ctx, errChan)
	go writer.Serve(ctx, writeQueueChan, errChan)

        return <-errChan
}
```

autholykos (changes requested):
LGTM, neatpicking on channel returning

cmd/dusk/server.go
julesdesmit: Just changed it to that :)

autholykos:
LGTM

PR #794, opened 2020-12-15
Remove error return value from SpawnReader

### LINKED ISSUE #750: Peer method NewReader returns error, but error is always nil. Return err is checked and will panic if not nil wich makes no sense since it is always nil
Method that should be updated, as we can see it will always return nil:
```
// NewReader returns a Reader. It will still need to be initialized by
// running ReadLoop in a goroutine.
func NewReader(conn net.Conn, gossip *processing.Gossip, dupeMap *dupemap.DupeMap, publisher eventbus.Publisher, rpcBus *rpcbus.RPCBus, counter *chainsync.Counter, responseChan chan<- *bytes.Buffer, exitChan chan<- struct{}) (*Reader, error) {
	pconn := &Connection{
		Conn:   conn,
		gossip: gossip,
	}

	_, db := heavy.CreateDBConnection()

	dataRequestor := responding.NewDataRequestor(db, rpcBus, responseChan)

	reader := &Reader{
		Connection: pconn,
		exitChan:   exitChan,
		router: &messageRouter{
			publisher:       publisher,
			dupeMap:         dupeMap,
			blockHashBroker: responding.NewBlockHashBroker(db, responseChan),
			synchronizer:    chainsync.NewChainSynchronizer(publisher, rpcBus, responseChan, counter),
			dataRequestor:   dataRequestor,
			dataBroker:      responding.NewDataBroker(db, rpcBus, responseChan),
			candidateBroker: responding.NewCandidateBroker(rpcBus, responseChan),
			ponger:          processing.NewPonger(responseChan),
			peerInfo:        conn.RemoteAddr().String(),
		},
	}

	// On each new connection the node sends topics.Mempool to retrieve mempool
	// txs from the new peer
	go func() {
		if err := dataRequestor.RequestMempoolItems(); err != nil {
			l.WithError(err).Warnln("error sending topics.Mempool message")
		}
	}()

	return reader, nil
}
```

Err check to be removed:
```
peerReader, err := peer.NewReader(conn, s.gossip, s.dupeMap, s.eventBus, s.rpcBus, s.counter, writeQueueChan, exitChan)
if err != nil {
	log.Panic(err)
}
```

Proposed changes:
Update method signature to not return error, and update err check to not do it anymore since it does not make any sense.

julesdesmit:
This value would always be nil, and thus is unnecessary.

Fixes #750

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -238,11 +238,7 @@ func Setup() *Server {
 // OnAccept read incoming packet from the peers
 func (s *Server) OnAccept(conn net.Conn) {
 	writeQueueChan := make(chan bytes.Buffer, 1000)
-	peerReader, err := s.readerFactory.SpawnReader(conn, s.gossip, s.dupeMap, writeQueueChan)
-	if err != nil {
-		panic(err)
-	}
-
+	peerReader := s.readerFactory.SpawnReader(conn, s.gossip, s.dupeMap, writeQueueChan)
 	if err := peerReader.Accept(); err != nil {
 		logServer.WithError(err).Warnln("OnAccept, problem performing handshake")
 		return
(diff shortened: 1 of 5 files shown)


autholykos (approved):
LGTM

PR #796, opened 2020-12-17
Change republishing to be in line with protocol

### LINKED ISSUE #758: Add verification logic to Republisher
Consensus steps are (likely) instantiated at each round while the message republishers are instantiated only once at consensus startup. Since for the most part, the republishers need to republish valid messages, they also need to access verification logic. This suggests that the verification functions should probably be extrapolated from the steps and passed to both the entities at different times if possible.

This would be especially recommended if verifications can be pure functions. Othewise we might considering some controlled duplication and slightly different verifications

### LINKED ISSUE #760: Refactor selection step
The logic for the selection is wasteful and could be improved.

1. Candidate should be verified only when the score received is greater than the `bestEvent`
2. Republishing should be handled by the selection directly. It does not make sense to republish messages associated to a lower score than the current one
3. Republishing should happened the score message is (sufficiently) verified. It might make sense to skip the score proof verification, but even in this case, the candidate verification must pass before a message is republished

### LINKED ISSUE #798: Synchronizer can get stuck if blocks are added mid-block-acceptance procedure
A hang was observed on the test-harness where the synchronizer would request missing blocks, receive the first one, and then exit. As this happened, concurrently, a successive block was received, stored and then left to it's own devices, as the block acceptance procedure still hadn't finished, and consequently, this block is never processed, causing a node to hang during syncs.

This can easily be mitigated by checking the block pool each time we receive a block. We may lose some time to this, but it will allow us to be fail-safe.

julesdesmit:
Since messages are only supposed to be
republished when the node is on the same state as the message,
we are now gossiping them directly out of the consensus components
after they are sufficiently verified. This also, currently, makes
the Republisher obsolete, but the code will be kept in case of
utility in future scenarios.

Fixes #760, fixes #758, fixes #798


autholykos (approved):
LGTM

PR #797, opened 2020-12-17
Add the candidate requestor to the Agreement

### LINKED ISSUE #770: Add Requestor at the end of the consensus if the Agreement converged on an unknown candidate
At the moment, the `Requestor` is only used by the `firststep.Phase` reduction. It might be that the `Agreement` would converge on a candidate we did not receive (in case of extreme network fragmentation). In this case we must check at the end of the loop if the candidate is known. If it is not, we need to use the `Requestor` to fetch it from the network

julesdesmit:
Fixes #770

### DIFF
--- a/pkg/core/consensus/agreement/step.go
+++ b/pkg/core/consensus/agreement/step.go
@@ -2,7 +2,9 @@ package agreement
 
 import (
 	"context"
+	"time"
 
+	"github.com/dusk-network/dusk-blockchain/pkg/core/candidate"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/data/block"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/database"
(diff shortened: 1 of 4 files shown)


pkg/core/consensus/agreement/step.go
```diff
@@ -116,20 +120,37 @@ func (s *Loop) shouldCollectNow(a message.Message, round uint64, queue *consensu
...
 
 	cm.Header.Certificate = cert
 	return cm, nil
 }
 
+func (s *Loop) requestCandidate(ctx context.Context, hash []byte) (block.Block, error) {
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(2*time.Second))
+	// Ensure we release the resources associated to this context.
+	defer cancel()
+	cm, err := s.requestor.RequestCandidate(ctx, hash)
```
autholykos, line 146: why not returning directly the call to the RequestCandidate? If there is an error, the caller will ignore the `cm` variable

autholykos (changes requested):

autholykos:
LGTM apart from a nit on a couple of redundant lines. Feel free to merge directly once that is taken care of

pkg/core/consensus/agreement/step.go
julesdesmit: Makes sense 👍

PR #800, opened 2020-12-18
Add a case for reading password from a non-terminal stdin

### LINKED ISSUE #799: Write a case that handles starting a node without a terminal
Currently, the password is read directly on startup using `terminal.ReadPassword`. This works fine for the normal user, but for automated deployments, this will not work, since the process stdin is not a TTY. If the process stdin is detected not to be a TTY, we should switch to using a `bufio.Scanner` to catch the password.

julesdesmit:
Fixes #799

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -1,11 +1,10 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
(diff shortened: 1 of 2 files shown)


autholykos:
LGTM

PR #802, opened 2020-12-18
Add code branches for Kadcasting consensus messages

### LINKED ISSUE #489: Propagate topics.Candidate in kadcast network

### LINKED ISSUE #801: Add Kadcast propagation for consensus messages

julesdesmit:
Fixes #801

### DIFF
--- a/pkg/core/consensus/selection/step.go
+++ b/pkg/core/consensus/selection/step.go
@@ -1,11 +1,13 @@
 package selection
 
 import (
+	"bytes"
 	"context"
 	"os"
 	"strconv"
 	"time"
 
+	"github.com/dusk-network/dusk-blockchain/pkg/config"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/candidate"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/blockgenerator"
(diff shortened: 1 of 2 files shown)


pkg/core/consensus/selection/step.go
```diff
@@ -95,6 +98,21 @@ func (p *Phase) generateCandidate(ctx context.Context, round consensus.RoundUpda
 		WithField("round", round.Round).
 		Debugln("sending score")
 
+	if config.Get().Kadcast.Enabled {
```
autholykos, line 101: I am not a fan of littering the code with configuration checks that should only matter at startup (once the networking method is established we don't ever look back). Maybe we should consider using an interface that gets injected by the server at startup to the consensus loop. Will open a new issue

autholykos (approved):
This can be merged keeping in mind that it is a temporary solution.

pkg/core/consensus/selection/step.go
julesdesmit: I had this same idea. I've held off on it for now since my focus shifted to devnet, but I was considering this as well.

autholykos:
Duplicate of #819

PR #812, opened 2020-12-19
Add license headers to all go files

### LINKED ISSUE #774: Add license headers to all source files
It would also be good, much like in the rust repositories, to check whether or not the proper license header has been added, and failing the CI if this is not the case.

julesdesmit:
Fixes #774


autholykos (approved):
LGTM

PR #814, opened 2020-12-21
Add a timeout context for establishing GRPC connections

### LINKED ISSUE #803: Connection to GRPC server should time out and panic on failure
There is currently no simple log entry that indicates a node gets stuck when the GRPC server is not online. To catch this issue more easily, we should add a timeout and a panic call on failure to connect, so that it is easier to troubleshoot.

julesdesmit:
Fixes #803

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -140,7 +140,9 @@ func Setup() *Server {
 
 	// Instantiate gRPC client
 	// TODO: get address from config
-	proxy, ruskConn := setupGRPCClients(ctx)
+	gctx, cancel := context.WithTimeout(ctx, time.Duration(cfg.Get().RPC.Rusk.ConnectionTimeout)*time.Millisecond)
+	defer cancel()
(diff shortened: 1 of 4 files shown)


cmd/dusk/server.go
```diff
@@ -140,7 +140,9 @@ func Setup() *Server {
 
 	// Instantiate gRPC client
 	// TODO: get address from config
-	proxy, ruskConn := setupGRPCClients(ctx)
+	gctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
autholykos, line 143: Maybe add this to a configuration?

autholykos (changes requested):

cmd/dusk/server.go
autholykos: Can be merged immediately afterwards

julesdesmit: Makes sense!

PR #818, opened 2020-12-21
Use context cancellation in mempool, and remove shared context in mempool testing package

### LINKED ISSUE #807: Reinstate the stake/bid automatons
Currently, they are skipped during setup, and should be started either immediately, or after user input from the wallet.

### LINKED ISSUE #809: Use context in Mempool, instead of a quitChan
This is more idiomatic, and in line with the rest of the codebase.

### LINKED ISSUE #817: Modify mempool testing file to avoid using a shared context
This shared context object often causes issues when amendments need to be made to the mempool, and hinders progress. Besides this, it strikes me as an anti-pattern to share a context object between unit tests, which should ideally be isolated and fine to be ran concurrently without usage of mutexes or anything of the sort.

julesdesmit:
Fixes #809, fixes #817

### DIFF
--- a/pkg/core/mempool/mempool.go
+++ b/pkg/core/mempool/mempool.go
@@ -63,23 +63,23 @@ type Mempool struct {
 
 	// the magic function that knows best what is valid chain Tx
 	verifier transactions.UnconfirmedTxProber
-	quitChan chan struct{}
-
-	ctx context.Context
 }
 
 // checkTx is responsible to determine if a tx is valid or not.
 // Among the other checks, the underlying verifier also checks double spending
 func (m *Mempool) checkTx(tx transactions.ContractCall) error {
+	ctx, cancel := context.WithTimeout(context.Background(),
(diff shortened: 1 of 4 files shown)


pkg/core/mempool/mempool.go
```diff
@@ -145,7 +143,7 @@ func (m *Mempool) Run() {
 			case <-time.After(20 * time.Second):
 				m.onIdle()
 			// Mempool terminating
-			case <-m.quitChan:
+			case <-m.ctx.Done():
```
autholykos, line 146: a better option than saving the context in the `Mempool` state would be to introduce `context.Context` as parameter of `Run` and let the server call it instead of implicitly being called by the constructor.

autholykos (changes requested):
Nitpicking on passing context. The rest looks good

pkg/core/mempool/mempool.go
julesdesmit: Yep, makes sense!

PR #821, opened 2020-12-21
Remove obsoleted TODO

julesdesmit:


julesdesmit:
This PR is not valid, as there has been a misunderstanding.

PR #829, opened 2020-12-22
Allow consensus to run without a block generator, populate consensus values on startup

### LINKED ISSUE #825: Allow nodes to run consensus if there are no bid values present
Currently, a node can not start the consensus loop if there are no bid values present in the DB. Since not all nodes will be block generators, this should be allowed - an error is currently returned and the loop is stopped instead.

### LINKED ISSUE #827: Populate bid values on startup, regardless of legacy flag
This allows devnet to run smoothly - as well as any other network. We need to have consensus values at bootstrap time in order to be able to create blocks.

### LINKED ISSUE #831: Genesis coinbases should include calldata
This avoids observed EOF errors in the rusk mock server.

julesdesmit:
Fixes #825, fixes #827, fixes #831


autholykos (approved):
LGTM

PR #842, opened 2020-12-30
Change 'trace message routing' log level to trace

### LINKED ISSUE #841: Change 'trace message routing' level to trace
Since we use the word trace in the log message, I think that's a sensible change to make. Moreover, this is polluting the logs massively on DevNet, and really doesn't belong in the debug level.

julesdesmit:
Fixes #841


autholykos (approved):
LGTM

PR #870, opened 2021-01-21
Start consensus only after all nodes have been initialized

### LINKED ISSUE #866: Stabilize TestConsensus
This is the biggest source of random CI failure at the moment, due to `block could not be received from the network` errors.

julesdesmit:
This fixes a bug, where some nodes would not
receive all messages properly, and cause TestConsensus to fail
over a panic when attempting to retrieve a candidate block
from the network. The panic occurs because there are no
CandidateBrokers initialized in this test, and thus the request
message falls on deaf ears. The upward propagation of the error
ultimately leads to a panic.

Fixes #866

### DIFF
--- a/pkg/core/consensus/testing/consensus_test.go
+++ b/pkg/core/consensus/testing/consensus_test.go
@@ -67,6 +67,15 @@ func TestConsensus(t *testing.T) {
 		defer cancelNode()
 	}
 
+	// Start the consensus loop on all nodes
+	for _, n := range nodes {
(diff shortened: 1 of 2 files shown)


goshawk-3 (approved):
LGTM

Looks stable:
```
go test -v --count=10 -run TestConsensus

--- PASS: TestConsensus (55.17s)
PASS
ok      github.com/dusk-network/dusk-blockchain/pkg/core/consensus/testing      552.452s
```

PR #871, opened 2021-01-21
Add step field to log entries for shouldProcess

### LINKED ISSUE #868: Add step field to log entries made by `ShouldProcess` in selection and reduction
To provide a developer with potentially useful info, we should also add the event step to log entries, when logging a discarded or queued event.

julesdesmit:
Fixes #868


goshawk-3 (approved):
LGTM

PR #879, opened 2021-01-26
Close websockets connections when they are not in use

### LINKED ISSUE #852: Close websockets connections when they are not in use
As per comment:

```
// TODO: it appears that the connection is not actually discarded. Maybe the GC
// finalizes it, but it should be checked what exactly happens to the
// connection if all brokers are busy
func (bp *BrokerPool) PushConn(conn *websocket.Conn) {
```

BrokerPool should close WS conn when the connection is discarded otherwise connection leakage could happen.

goshawk-3:
fixes #852


julesdesmit (approved):
LGTM

PR #885, opened 2021-01-28
 Convert StandardType (3) to TransferType (0)

### LINKED ISSUE #880: Debug network stalling after sending transfer transaction
Immediately after sending grpc call Transfer to all 9 nodes of test-harness, network consensus gets stalled. This might be due to a bug with mocking services but still needs to be investigated.

goshawk-3:
fixes #880

### DIFF
--- a/harness/engine/profiles.go
+++ b/harness/engine/profiles.go
@@ -64,7 +64,7 @@ func Profile1(index int, node *DuskNode, walletPath string) {
 	viper.Set("timeout.timeoutkeepalivetime", 30)
 
 	viper.Set("logger.output", node.Dir+"/dusk")
-	viper.Set("logger.level", "warn")
+	viper.Set("logger.level", "info")
 
 	viper.Set("gql.address", node.Cfg.Gql.Address)
 	viper.Set("gql.network", node.Cfg.Gql.Network)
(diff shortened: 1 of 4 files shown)


harness/engine/profiles.go
```diff
@@ -64,7 +64,7 @@ func Profile1(index int, node *DuskNode, walletPath string) {
 	viper.Set("timeout.timeoutkeepalivetime", 30)
 
 	viper.Set("logger.output", node.Dir+"/dusk")
-	viper.Set("logger.level", "warn")
+	viper.Set("logger.level", "info")
```
julesdesmit, line 67: Is it not better to have more granular logging levels for the test-harness, in case we need to debug?

julesdesmit (changes requested):
Just one question.

harness/engine/profiles.go
goshawk-3: For this PR  it should be fine to change it from `warn` to `info`. For debugging test-harness problems within CI, `debug` level would be must but let's address in the next PR.

julesdesmit: I don't understand - what is the point of changing it, and why should we open a new PR to change it back? A change like this may have made sense for debugging during this PR, but it doesn't seem like a change that should be permanent.

goshawk-3: Agreed.  Change removed.

julesdesmit (approved):
LGTM

PR #897, opened 2021-02-03
Reintroduce synchronicity between the Synchronizer, Chain and Consensus loop

### LINKED ISSUE #893: Reintroduce synchronicity between the Synchronizer, Chain and Consensus loop
Since we have noticed an increasing number of bugs with the current consensus pipeline setup, I think it is sensible to revert back to a more synchronized approach, as introduced in the last architectural refactor. The aim is to maintain the separation of concerns between the Chain and Synchronizer, but to re-introduce the model where goroutines replace each other as the consensus loop, instead of maintaining one goroutine for as long as possible.

This should help avoid stalls and potential bugs.

julesdesmit:
Fixes #893

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -60,8 +60,8 @@ type Loader interface {
 
 // Ledger is the Chain interface used in tests.
 type Ledger interface {
-	ProcessSuccessiveBlock(block.Block) error
-	ProcessSyncBlock(block.Block) error
+	TryNextConsecutiveBlockInSync(block.Block) error
+	TryNextConsecutiveBlockOutSync(block.Block) error
 	ProduceBlock() error
 	StopBlockProduction()
 }
@@ -154,10 +154,10 @@ func (c *Chain) StopBlockProduction() {
 	select {
 	case c.stopConsensusChan <- struct{}{}:
 	// If there is nobody listening on the other end, it could very well be that
-	// `crunchBlocks` is attempting to take control of the mutex. In this instance,
-	// we can forego the channel send here, as the release of the mutex will result
-	// in a bump to the tip height, making the consensus results obsolete, and
-	// ending the lifetime of that goroutine.
+	// `acceptConsensusResults` is attempting to take control of the mutex.
+	// In this instance, we can forego the channel send here, as the release of
+	// the mutex will result in a bump to the tip height, making the consensus
+	// results obsolete, and ending the lifetime of that goroutine.
 	default:
 	}
 }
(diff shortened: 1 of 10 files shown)


pkg/core/chain/chain.go
```diff
@@ -143,111 +211,129 @@ func New(ctx context.Context, db database.DB, eventBus *eventbus.EventBus, rpcBu
...
-	select {
-	case c.CatchBlockChan <- consensus.Results{Blk: blk, Err: nil}:
-	default:
-		return errors.New("block discarded")
+func (c *Chain) ProcessSuccessiveBlock(blk block.Block) error {
+	// If we can not stop the consensus, it means that the ProduceBlock goroutine
+	// is waiting to take ownership of the mutex. In this scenario, it means
+	// that consensus has converged on a block. We will give precendence to
+	// self-finalized blocks over network-received ones, so we relinquish control
+	// over the mutex in case of an error.
```
autholykos, line 331: The description does not really reflect what is in the code. I cannot easily see where the control over the mutex gets relinquished as a consequence of errors returned by `StopBlockProduction`. If a refactoring leading to more explicit code is not easy to obtain, at the very least the comment should be more specific and detailed about side effects in the mutex handling (which I suppose should be in the synchronization state-machine, but I am not sure)

autholykos (changes requested):
I like the directionf of the refactoring, but I think some improvements can be made in making the lock handling more maintainable/readable

pkg/core/chain/chain.go
```diff
@@ -60,11 +60,10 @@ type Loader interface {
 
 // Ledger is the Chain interface used in tests.
 type Ledger interface {
-	CurrentHeight() uint64
-	ProcessSucceedingBlock(block.Block) error
+	ProcessSuccessiveBlock(block.Block) error
```
autholykos, line 63: Can we rename this into `TryNextConsecutiveBlockInSync`? `Successive` is quite an odd adjective

pkg/core/chain/chain.go
```diff
@@ -143,113 +148,147 @@ func New(ctx context.Context, db database.DB, eventBus *eventbus.EventBus, rpcBu
...
 
+	go func(ctx context.Context, cancel context.CancelFunc, winnerChan chan consensus.Results) {
+		defer cancel()
+		c.crunchBlocks(ctx, winnerChan)
+	}(ctx, cancel, winnerChan)
+
+	return nil
+}
+
+func (c *Chain) crunchBlocks(ctx context.Context, winnerChan chan consensus.Results) {
```
autholykos, line 213: rename into `acceptConsensusResults`

pkg/core/chain/chain.go
```diff
@@ -60,11 +60,10 @@ type Loader interface {
 
 // Ledger is the Chain interface used in tests.
 type Ledger interface {
-	CurrentHeight() uint64
-	ProcessSucceedingBlock(block.Block) error
+	ProcessSuccessiveBlock(block.Block) error
 	ProcessSyncBlock(block.Block) error
```
autholykos, line 64: Specularly, this should be called `TryNextConsecutiveBlockOutSync`

autholykos (changes requested):
I think the refactoring makes a lot of sense. I just left some renaming suggestions for extra-clarity

autholykos (approved):
LGTM (good one on removing the goroutine wrapper, I missed that one)

PR #903, opened 2021-02-05
Move GetSyncProgress method back to the Chain

### LINKED ISSUE #899: Move GetSyncProgress method back to the Chain
As a result of #897, the synchronizer is no longer a stand-alone component, and should not register as a GRPC server. Furthermore, as it lacks access to the current height and highest seen variables, it should no longer divulge this information. The chain will be the best candidate now.

julesdesmit:
Fixes #899


autholykos (approved):
LGTM

PR #904, opened 2021-02-06
Restrict outgoing stake transactions if node is not synced

### LINKED ISSUE #875: Restrict node's ability to transact until it is fully synced
As discovered in #855, the DevNet stalled after introducing too many nodes at the same time. This is because they were sending fully valid transactions from the moment they started, and since they weren't yet synced, entered the committee, but could not participate in consensus.

In real-world scenarios, it is most likely the case that a node will not be able to transact until they are fully synced anyway, as it is not standard for nodes to start up with any amount of DUSK. However, it is still sensible to disallow transacting until the node is fully synced, as a node which was stopped and started might not have a proper view of the UTXO set, and could submit potentially invalid transactions, or even re-enact the scenario described above.

julesdesmit:
Fixes #875

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -463,14 +463,15 @@ func (c *Chain) getRoundUpdate() consensus.RoundUpdate {
 
 // GetSyncProgress returns how close the node is to being synced to the tip,
 // as a percentage value.
-// NOTE: this is just here to satisfy the grpc interface. It should be removed
-// and the method should be moved to a synchronizer service.
 func (c *Chain) GetSyncProgress(_ context.Context, e *node.EmptyRequest) (*node.SyncProgressResponse, error) {
-	return &node.SyncProgressResponse{Progress: float32(100.0)}, nil
+	return &node.SyncProgressResponse{Progress: float32(c.CalculateSyncProgress())}, nil
 }
 
 // CalculateSyncProgress of the node.
 func (c *Chain) CalculateSyncProgress() float64 {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
 	if c.highestSeen == 0 {
 		return 0.0
 	}
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -25,7 +25,6 @@ import (
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/topics"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/diagnostics"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/eventbus"
-	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/rpcbus"
 	"github.com/dusk-network/dusk-protobuf/autogen/go/node"
 	logger "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
@@ -70,7 +69,6 @@ type Ledger interface {
 // This struct will be aware of the current state of the node.
 type Chain struct {
 	eventBus *eventbus.EventBus
-	rpcBus   *rpcbus.RPCBus
 	db       database.DB
 
 	// loader abstracts away the persistence aspect of Block operations.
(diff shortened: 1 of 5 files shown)


pkg/core/chain/chain.go
```diff
@@ -473,6 +469,20 @@ func (c *Chain) GetSyncProgress(_ context.Context, e *node.EmptyRequest) (*node.
 	return &node.SyncProgressResponse{Progress: float32(100.0)}, nil
 }
 
+// CalculateSyncProgress of the node.
+func (c *Chain) CalculateSyncProgress() float64 {
+	if c.highestSeen == 0 {
```
autholykos, line 474: since this method is now used within the `Transactor`, shouldn't `c.highestSeen` and `c.tip` be protected by a read-lock?

pkg/core/transactor/transactor.go
```diff
@@ -81,6 +85,11 @@ func (t *Transactor) Listen() {
 	for {
 		select {
 		case r := <-t.stakeChan:
+			// Are we synced?
+			if t.getSyncProgress() != 100 {
+				continue
+			}
```
autholykos, line 91: if we are not synced, what happens to the attempted stake transaction? Does it just get garbage collected and forgotten? I don't see a _retry_ mechanism

autholykos (changes requested):

pkg/core/chain/chain.go
julesdesmit: Good catch, it is definitely needed

pkg/core/transactor/transactor.go
julesdesmit: You're absolutely right, forgot to add it. Will do so now

pkg/core/transactor/transactor.go
```diff
@@ -105,6 +115,27 @@ func (t *Transactor) Listen() {
...
+	if t.getSyncProgress() == 100 {
+		return true
+	}
+
+	// Check for our sync progress three more times. If we still can't stake after
+	// the third check, it's better to just return false, and give the Transactor
+	// control back over the Listen goroutine.
+	interval := 5 * time.Second
+
+	for i := 0; i < 3; i++ {
```
goshawk-3, line 128: When I suggested to have a retry mechanism, I meant that it could be useful at the grpc client-side mostly (e.g wallet cli) where retry period can be also configurable and we will not block the entire Transactor on each stake Tx. But I believe server-side mechanism could be good for now too.

pkg/core/transactor/transactor.go
```diff
@@ -114,6 +115,27 @@ func (t *Transactor) Listen() {
...
+		time.Sleep(interval * time.Duration(i+1))
+
+		if t.getSyncProgress() == 100 {
+			return true
+		}
+	}
```
autholykos, line 134: I think that this kind of retry is too naìve. It is fine to perform a number of attempts, but the wallet needs to get  feeback rather allow for silent failure. It might make sense to use the RPCBus here and return feedback to the wallet

autholykos (changes requested):

pkg/core/transactor/transactor.go
julesdesmit: Well, using the RPCBus here to fetch the sync progress really doesn't change much, besides the infrastructure we're using (unless I'm missing something). I can of course add logging, though.

goshawk-3 (approved):

pkg/core/transactor/transactor.go
julesdesmit: The reason I opted for a server-side mechanism is because the grpc client is not the only one that can ask the transactor to stake (the `StakeAutomaton` does too). But, your point does remind me that the retry is therefore in the wrong place, as it needs to be touched by both code paths.

pkg/core/transactor/transactor.go
```diff
@@ -163,6 +158,11 @@ func (t *Transactor) Bid(ctx context.Context, c *node.BidRequest) (*node.Transac
 
 // Stake will create a staking transaction.
 func (t *Transactor) Stake(ctx context.Context, c *node.StakeRequest) (*node.TransactionResponse, error) {
+	// Are we synced?
+	if !t.canStake() {
+		return nil, errors.New("node is not synced")
+	}
+
```
autholykos, line 171: I like it. Failing early before pushing requests in the channel is a good strategy

autholykos (approved):
LGTM

pkg/core/transactor/transactor.go
autholykos: Make sure to add this to the development guidelines

PR #905, opened 2021-02-06
Add a grace period for new stakes to become active

### LINKED ISSUE #901: Add a grace period for Stakes to become active
Currently, when not using Rusk, stakes will become active immediately as they are added to the blockchain. This does not line up with the protocol specifications, and should be implemented on the Golang side too.

julesdesmit:
Fixes #901


autholykos (approved):
LGTM

PR #909, opened 2021-02-09
Secure `dupemap` against spam attacks

### LINKED ISSUE #200: Secure `dupemap` against spam attacks
The node uses the `dupemap` to prevent processing message duplicates. Such cache does not have a spam protection mechanism to prevent an attacker to fill up the memory of the node.

goshawk-3:
fixes #200


julesdesmit (approved):
LGTM!

PR #948, opened 2021-02-23
Introduce PeerID to all MessageProcessor handlers

### LINKED ISSUE #947: Implement PeerID parameter for all MessageProcessor handlers
PeerID (a.k.a peer IPv4/IPv6 address) should be passed to the upper layers in order to allow implementing custom logic based on the source of the message. This would be useful mostly in situations when a Peer is dishonest or misbehaves.

E.g Synchronizer needs to know the address of the Peer that is sending forged blocks (see also: #766)

*Pseudo code*:

```go
func (m *MessageProcessor) process(peerID string, msg message.Message, respChan chan<- bytes.Buffer) error {

...
bufs, err := processFn(peerID, msg)
	if err != nil {
		return err
	}
}

...
func (c *Chain) ProcessBlockFromNetwork(peerID string, m message.Message) {

...
}



```

goshawk-3:
fixes #947

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ go-analyzer: ## Run go-analyzer
 	GOBIN=$(PWD)/bin go run scripts/build.go go-analyzer
 test: ## Run unittests
 	@go test $(TFLAGS) -short ${PKG_LIST}
+test-harness-unit: build ## Run a specified harness unit test e.g make UNIT_TEST=TestForgedBlock test-harness-unit
+	MOCK_ADDRESS=127.0.0.1:8080 DUSK_NETWORK_SIZE=9 DUSK_BLOCKCHAIN=${PWD}/bin/dusk DUSK_UTILS=${PWD}/bin/utils DUSK_SEEDER=${PWD}/bin/voucher DUSK_WALLET_PASS="password" \
+	go test -v --count=1 --test.timeout=0 ./harness/tests/ -run $(UNIT_TEST) -args -enable
 test-harness: ## Run harness tests
 	@go test -v --count=1 --test.timeout=0 ./harness/tests/ -args -enable
 test-harness-ci: build
--- a/pkg/core/candidate/requestor.go
+++ b/pkg/core/candidate/requestor.go
@@ -41,7 +41,7 @@ func NewRequestor(publisher eventbus.Publisher) *Requestor {
 
 // ProcessCandidate will process a received Candidate message.
 // Invalid and non-matching Candidate messages are discarded.
-func (r *Requestor) ProcessCandidate(msg message.Message) ([]bytes.Buffer, error) {
+func (r *Requestor) ProcessCandidate(srcPeerID string, msg message.Message) ([]bytes.Buffer, error) {
 	if r.isRequesting() {
 		if err := Validate(msg); err != nil {
 			return nil, err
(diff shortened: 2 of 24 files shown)


julesdesmit (changes requested):
I don't really understand the point of this PR - it seems you are polluting the `peer.ProcessorFunc` interface with an argument that it doesn't use anywhere. Even if you were planning on using it later, does it not make sense to simply just keep the `PeerID` contained in the [goroutine where we call `processor.Collect`](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/p2p/peer/peer.go#L371), so that you can use it for logging on potential errors.

If I understood correctly from your previous PR, you were intending on using this `PeerID` in order to punish peers that attempt to stall consensus by forging blocks and gossiping them. Since this concerns a punishment on the peer layer, I don't think this information has any business to go outside of the peer layer.

I hope it makes sense. If I have misunderstood something, please let me know.

goshawk-3:
> I don't really understand the point of this PR - it seems you are polluting the `peer.ProcessorFunc` interface with an argument that it doesn't use anywhere. Even if you were planning on using it later, does it not make sense to simply just keep the `PeerID` contained in the [goroutine where we call `processor.Collect`](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/p2p/peer/peer.go#L371), so that you can use it for logging on potential errors?
> 
> If I understood correctly from your previous PR, you were intending on using this `PeerID` in order to punish peers that attempt to stall consensus by forging blocks and gossiping them. Since this concerns a punishment on the peer layer, I don't think this information has any business going outside of the peer layer - you possibly don't even have to quantify the `PeerID` at all, because when an error is returned that the peer needs to be punished for, the goroutine has direct access to the `Peer` that spawned it.
> 
> I hope it makes sense. If I have misunderstood something, please let me know.

Thanks for your comments, @julesdesmit . 
I think we are not on the same page. In short, the `srcPeerID` is used by `outSyncTimer` on resetting the timer. `outSyncTimer` must ensure that the `Peer` that starts syncing is the one that is allowed to reset the timer.

But much better to jump into a call to clarify both opinions.

See also: https://github.com/dusk-network/dusk-blockchain/blob/26d901767be2218fa8ed7d3a6b04db0687234079/pkg/core/chain/synchronizer.go#L78

julesdesmit:
>Thanks for your comments, @julesdesmit .
I think we are not on the same page. In short, the srcPeerID is used by outSyncTimer on resetting the timer. outSyncTimer must ensure that the Peer that starts syncing is the one that is allowed to reset the timer.

>But much better to jump into a call to clarify both opinions.

I see - I was unaware of this particular part of the `outSyncTimer`. In that case, you are right, and I believe it is fine to merge - however, I would like to explore a possible re-design in the future, as I feel like needing to adjust a widely-used interface for a single component that implements it, is a bit of a code smell.

julesdesmit (approved):

goshawk-3:
> > Thanks for your comments, @julesdesmit .
> > I think we are not on the same page. In short, the srcPeerID is used by outSyncTimer on resetting the timer. outSyncTimer must ensure that the Peer that starts syncing is the one that is allowed to reset the timer.
> 
> > But much better to jump into a call to clarify both opinions.
> 
> I see - I was unaware of this particular part of the `outSyncTimer`. In that case, you are right, and I believe it is fine to merge - however, I would like to explore a possible re-design in the future, as I feel like needing to adjust a widely-used interface for a single component that implements it, is a bit of a code smell.

Thanks! Agreed we can discuss this further later on together with `banscore` implementation.

PR #972, opened 2021-03-16
Introduce TTL property to Cucko Filter instance per round

### LINKED ISSUE #971: Handle an edge case where a Node plays dishonestly in synchronization process
A network node might intentionally or accidentally fail to respond to wire messages `topics.GetData` or `topics.GetBlocks`. In these cases, the Node requestor (sender of `topics.GetData` or `topics.GetBlocks`) might get into wrong state where `dupemap` rejects a block hash infinitely.

The result of such sequence of events would leave the Node requestor in `livelock`

Observed in `test-harness` when one of the nodes is dishonest.
```
NODE:9500:
{"data":{"blocks":[{"header":{"height":32}}]}}
NODE:9501:
{"data":{"blocks":[{"header":{"height":32}}]}}
NODE:9502:
{"data":{"blocks":[{"header":{"height":32}}]}}
NODE:9503:
{"data":{"blocks":[{"header":{"height":3}}]}}

```

goshawk-3:
fixes #971

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -359,7 +359,7 @@ func (s *Server) Close() {
 func registerPeerServices(processor *peer.MessageProcessor, db database.DB, eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus) {
 	processor.Register(topics.Ping, responding.ProcessPing)
 	dataBroker := responding.NewDataBroker(db, rpcBus)
-	dataRequestor := responding.NewDataRequestor(db, rpcBus, eventBus)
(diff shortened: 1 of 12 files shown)


julesdesmit (changes requested):
I think this is quite an awesome re-design, well done! I believe there should be no repercussions going from event-based to quasi timer-based entry deletion. One thing I will request, however, is that the legacy code gets deleted. There is still leftover code relating to the `TmpMap` height, tolerance and certain cleaning functions which might now be obsolete, since we delete stuff on expiry, and not on reaching a certain height.

julesdesmit (approved):
LGTM!

PR #978, opened 2021-03-22
Add gas limit and gas price fields to tx queries

### LINKED ISSUE #977: Expose transaction gas information over GQL

julesdesmit:
Fixes #977

### DIFF
--- a/pkg/core/data/ipc/transactions/transaction.go
+++ b/pkg/core/data/ipc/transactions/transaction.go
@@ -188,9 +188,9 @@ func (t Transaction) Obfuscated() bool {
 }
 
 // Values returns the amount and fee spent in this transaction.
-// TODO: implement.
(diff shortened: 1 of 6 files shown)


pkg/core/data/ipc/transactions/transaction.go
```diff
@@ -190,7 +190,7 @@ func (t Transaction) Obfuscated() bool {
 // Values returns the amount and fee spent in this transaction.
 // TODO: implement.
```
autholykos, line 191: remove the TODO ;P

autholykos (approved):
LGTM

pkg/core/data/ipc/transactions/transaction.go
julesdesmit: It's still not entirely implemented, that's why I left it 😅

PR #985, opened 2021-03-23
Allow for easy configuration of genesis blocks, add a configuration for stress testing

### LINKED ISSUE #983: Regenerate a devnet genesis block which gives all participants many outputs
This is needed to properly stress test transaction capabilities in the network.

julesdesmit:
This allows us to easily perform transaction stress-testing.

This commit also changes the way genesis blocks are read, as it
was undesirable to have such a massive string in one of the source
files.

Fixes #983


harness/engine/node.go
```diff
@@ -38,6 +38,7 @@ func NewDuskNode(graphqlPort, nodeID int, profileID string, requireSession bool)
 
 	node.Cfg = config.Registry{}
 	node.Cfg.General.TestHarness = true
+	node.Cfg.General.GenesisPath = "$HOME/go/src/github.com/dusk-network/dusk-blockchain/devnetgenesis.txt"
```
goshawk-3, line 41: Suggestion: Do you think we can use relative paths instead? Similar approach is already used for `devnet-wallets//wallet0.dat"`.

As per https://github.com/dusk-network/dusk-blockchain/blob/53deecea47bb861fa1dcd2965c76f0d564ea1a8f/harness/engine/network.go#L233

```
[wallet]
  file = "/home/tech/repo/dusknetwork/dusk-blockchain/harness/tests/../../devnet-wallets//wallet0.dat"
```

harness/engine/profiles.go
```diff
@@ -30,6 +30,7 @@ func Profile1(index int, node *DuskNode, walletPath string) {
 	viper.Set("general.walletonly", "false")
 	viper.Set("general.safecallbacklistener", "false")
 	viper.Set("general.testharness", "true")
+	viper.Set("general.genesispath", "$HOME/go/src/github.com/dusk-network/dusk-blockchain/devnetgenesis.txt")
```
goshawk-3, line 33: devnetgenesis.txt seems not to be a text file? If not, could we use `dat` or `bin` extention?

harness/engine/node.go
julesdesmit: This isn't feasible due to tests - since when running tests, the working directory is dynamic and thus the relative path can change. I've opted for a different approach as I mentioned on the standup, which generates it deterministically on startup. This should relieve some of the portability issues as well on the CI runner and potentially on devnet.

goshawk-3: The relative path is relative to the repository directory not to the working directory. Otherwise, `devnet-wallets` would not work.

julesdesmit: `devnet-wallets` works because we always refer to it from the same, singular point. In the code block you posted you also show that the path is relative to the working directory within the harness package, `tests`.

goshawk-3: Might be missing something.
Please, use at least "$GOPATH"  instead of relying on its default location.

julesdesmit: Like I said, I am taking a different approach, so this will not be needed. It would possibly be better, though the runners do not work out of the $GOPATH either, as far as I'm aware. So that still would not work.

goshawk-3 (approved):

PR #993, opened 2021-03-29
Measure network TPS 

### LINKED ISSUE #938: Implement a test which can be ran by the test-harness to perform stress testing for transaction processing on local networks

goshawk-3:
fixes #938

### DIFF
--- a/harness/engine/utils.go
+++ b/harness/engine/utils.go
@@ -496,14 +496,12 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
 		return nil, err
 	}
 
-	if logrus.GetLevel() == logrus.DebugLevel {
-		logrus.
-			WithField("s_wallet", senderAddr).
-			WithField("r_wallet", recvAddr).
-			WithField("amount", amount).
-			WithField("fee", fee).
-			Debug("Sending transfer")
-	}
+	logrus.
+		WithField("s_wallet", senderAddr).
+		WithField("r_wallet", recvAddr).
+		WithField("amount", amount).
+		WithField("fee", fee).
+		Debug("Sending transfer")
 
 	// Send Transfer grpc command
 	c := n.grpcClients[n.nodes[senderNodeInd].Id]
@@ -547,14 +545,12 @@ func (n *Network) BatchSendTransferTx(t *testing.T, senderNodeInd uint, batchSiz
 		return err
 	}
 
-	if logrus.GetLevel() == logrus.DebugLevel {
-		logrus.
-			WithField("s_wallet", senderAddr).
-			WithField("r_wallet", recvAddr).
-			WithField("amount", amount).
-			WithField("fee", fee).
-			Debug("Sending transfer")
-	}
+	logrus.
+		WithField("s_wallet", senderAddr).
+		WithField("r_wallet", recvAddr).
+		WithField("amount", amount).
+		WithField("fee", fee).
+		Debug("Sending transfer")
 
 	// Send Transfer grpc command
 	c := n.grpcClients[n.nodes[senderNodeInd].Id]
@@ -571,15 +567,16 @@ func (n *Network) BatchSendTransferTx(t *testing.T, senderNodeInd uint, batchSiz
 	client := pb.NewTransactorClient(conn)
 
 	for i := uint(0); i < batchSize; i++ {
-		ctx, cancel := context.WithTimeout(context.Background(), timeout)
-		defer cancel()
-
 		req := pb.TransferRequest{Amount: amount, Address: pubKey, Fee: fee}
+
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
 
 		_, err := client.Transfer(ctx, &req)
 		if err != nil {
 			logrus.WithField("sender_index", senderNodeInd).Error(err)
 		}
+
+		cancel()
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


harness/engine/utils.go
```diff
@@ -452,12 +496,14 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
 		return nil, err
 	}
 
-	logrus.
-		WithField("s_wallet", senderAddr).
-		WithField("r_wallet", recvAddr).
-		WithField("amount", amount).
-		WithField("fee", fee).
-		Info("Sending transfer")
+	if logrus.GetLevel() == logrus.DebugLevel {
```
julesdesmit, line 499: This conditional seems a bit redundant - if we're not on `Debug` level, the entry will not get written.

harness/engine/utils.go
```diff
@@ -481,3 +527,84 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
...
+		return err
+	}
+
+	// Get wallet address of receiver
+	recvAddr, pubKey, err := n.GetWalletAddress(recvNodeInd)
+	if err != nil {
+		return err
+	}
+
+	if logrus.GetLevel() == logrus.DebugLevel {
```
julesdesmit, line 550: Same here

harness/engine/utils.go
```diff
@@ -481,3 +527,84 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
...
+	nodeInd := uint(0)
+
+	for {
+		time.Sleep(delay)
+
+		h, tps, blockTime, lastTxsCount, err := n.CalculateTPS(nodeInd)
+
+		if err == nil {
+			logrus.WithField("height", h).Infof("Measured TPS %.2f (%d/%d)", tps, lastTxsCount, blockTime)
+		} else {
```
julesdesmit, line 600: In accordance with the coding conventions, let's try and avoid using `else` whenever possible; this is because we try to avoid extra indentations where it's not needed.

You could rewrite it as such:

```golang
if err != nil {
    logrus.WithError(err).Warn("calculate tps failed")

    // Failover to another node if this fails
    nodeInd++
    if nodeInd >= uint(n.Size()) {
        nodeInd = 0
    }

    continue
}

logrus.WithField("height", h).Infof("Measured TPS %.2f (%d/%d)", tps, lastTxsCount, blockTime)
```

harness/engine/utils.go
```diff
@@ -481,3 +527,84 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
...
+// MonitorTPS tries to measure network TPS each N seconds.
+func (n *Network) MonitorTPS(delay time.Duration) {
+	// Start monitoring TPS value of the network
+	nodeInd := uint(0)
+
+	for {
+		time.Sleep(delay)
+
+		h, tps, blockTime, lastTxsCount, err := n.CalculateTPS(nodeInd)
+
```
julesdesmit, line 597: No need for an empty line here.

harness/engine/utils.go
```diff
@@ -481,3 +527,84 @@ func (n *Network) SendTransferTxCmd(senderNodeInd, recvNodeInd uint, amount, fee
...
+		if err == nil {
+			logrus.WithField("height", h).Infof("Measured TPS %.2f (%d/%d)", tps, lastTxsCount, blockTime)
+		} else {
+			logrus.WithError(err).Warn("calculate tps failed")
```
julesdesmit, line 601: Might be useful as well for the reader to include in this log entry that we're failing over to another node.

julesdesmit (changes requested):
Couple of style nits. The code itself looks great!

goshawk-3:
Thanks for the comments, @julesdesmit. A minor suggestion for any future review:

Apart from styling errors (which are good to improve for readability), please let's not forget to look for design flaws. Ya, I'm making design flaws sometimes and not only styling errors : ).

E.g  In this PR  `defer cancel()` is wrongly called in a loop which makes `timeout` property in erroneous. A comment about this one would be much more helpful than a comment about wrong `whitespace`.

julesdesmit:
Of course, I must've missed this in my review. I try to look for all kinds of errors, but only managed to catch styling errors this time.

julesdesmit:
Furthermore though @goshawk-3, when you create a context with a timeout, you always need to `defer cancel()` so that you clean up the resources associated with that context in all cases. Since you create a new context for each loop iteration, calling `defer cancel()` for each loop iteration is, in my view, correct behaviour. Unless I'm missing something :)

goshawk-3:
> Furthermore though @goshawk-3, when you create a context with a timeout, you always need to `defer cancel()` so that you clean up the resources associated with that context in all cases. Since you create a new context for each loop iteration, calling `defer cancel()` for each loop iteration is, in my view, correct behaviour. Unless I'm missing something :)

This is in a loop, so it should be called like:

```diff
-               ctx, cancel := context.WithTimeout(context.Background(), timeout)
-               defer cancel()
-
                req := pb.TransferRequest{Amount: amount, Address: pubKey, Fee: fee}
 
+               ctx, cancel := context.WithTimeout(context.Background(), timeout)
+
                _, err := client.Transfer(ctx, &req)
                if err != nil {
                        logrus.WithField("sender_index", senderNodeInd).Error(err)
                }
+
+               cancel()
```

julesdesmit:
@goshawk-3 Ahh, now I get you. Fair point!

julesdesmit (approved):
LGTM! Fine to merge once the CI passes.

PR #996, opened 2021-03-30
Add a utility which continually spams transactions from a node

### LINKED ISSUE #939: Create a stand-alone utility which can be generically used for transaction stress-testing
The idea behind this utility is that is should be pluggable anywhere (from unit tests to production environments) in order to "spam" the network with lots of transactions. The aim, is to find the upper limit of transaction throughput, and through heavy load, expose vulnerabilities or inefficiencies in the design of components in the node which process transactions (RUSK-related components excluded for now).

The fix for this issue should come with documentation and diagrams, on top of code.

julesdesmit:
Fixes #939

### DIFF
--- a/cmd/utils/tps/tps.go
+++ b/cmd/utils/tps/tps.go
@@ -15,6 +15,11 @@ import (
 	"google.golang.org/grpc"
 )
 
+const (
+	retryAmount = 5
+	retryDelay  = 5 * time.Second
+)
+
 // StartSpamming transactions from the given GRPC address. The utility will spam
 // Transfer transactions right back to the wallet it's coming from.
 func StartSpamming(addr string, delay int, amount uint64) error {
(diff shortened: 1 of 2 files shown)


cmd/utils/tps/tps.go
```diff
@@ -0,0 +1,108 @@
...
+func StartSpamming(addr string, delay int, amount uint64) error {
+	if addr == "" {
+		return errors.New("no GRPC address provided")
+	}
+
+	if amount == 0 {
+		amount = defaultAmount
+	}
+
+	conn, err := grpc.Dial(addr, grpc.WithInsecure())
```
goshawk-3, line 36: Could we add here `DialTimeout`  support? 
Also, we might need `grpc.WithBlock()` option.

julesdesmit: Good catch, will add!

goshawk-3 (approved):
LGTM

PR #1010, opened 2021-04-02
Introduce CPU profiling for Rusk Mock  and transactor stats

### LINKED ISSUE #1009: Enable CPU profiling for Rusk mock service

goshawk-3:
fixes #1009


pkg/core/transactor/listener.go
```diff
@@ -228,6 +230,10 @@ func (t *Transactor) handleSendStandardTx(req *node.TransferRequest) (*node.Tran
 		return nil, err
 	}
 
+	t.stats.transferCumulativeTime += (time.Now().UnixNano() - start)
```
julesdesmit, line 233: I think this stuff belongs more to the test harness. It's code related to integration testing/stress testing, and thus doesn't really belong in the node source.

pkg/core/transactor/transactor.go
```diff
@@ -23,6 +23,11 @@ import (
 	"google.golang.org/grpc"
 )
 
+type transactorStats struct {
```
julesdesmit, line 26: Ditto

julesdesmit (changes requested):

pkg/core/transactor/listener.go
goshawk-3: It's normal for a node to collect stats internally, especially if it needs to monitor external services behavior. This could be also exported to Grafana.  We definitely would need it when Rusk is integrated.

julesdesmit: It is fine to log certain stats, but I don't know how I feel about adding a struct to the transactor which holds cumulative information. I think logging time spent is fine, but then I think the metrics process should extrapolate this information to contain what you have added here, instead of putting it directly on the component.

I may be wrong, but it doesn't feel like the right place. Maybe @autholykos could weigh in?

goshawk-3: Agreed. This should be enabled in `debug` mode only. I'll keep only the timestamp logging. Later on, we may enrich monitoring capabilities with new stats, if needed.

julesdesmit (approved):
LGTM

PR #1018, opened 2021-04-14
Enable unix transport for Rusk service

### LINKED ISSUE #1016: Evaluate transaction throughput in a realistic gossip network
The output of this task should be a report of average TPS a network can achieve when:

- Gossip network is used for transaction propagation
- Network consists of at least 30 nodes
- Average network latency is 25ms

goshawk-3:
fixes #1016


julesdesmit (approved):
LGTM

PR #1022, opened 2021-04-19
Introduce Reserved field in Gossip structure

### LINKED ISSUE #1019: Investigate reasons behind block time being impacted by transactions propagation
As per finding from issue #1016, we should find out why consensus time increases in direct proportion with the count of transfer transactions sent to the network.

> 
Based on the results from the three tests above, it turns out that in a network with latency of 50ms gossiping a batch of transactions (200*20=4000)  has a noticeable impact upon average block time.  The average block time increases from 7 to 14 seconds.
NB. All tests were performed with `PassStateTransition=True` as a workaround for the significantly delayed `ExecuteStateTransition` response when batch of transactions is being sent.

goshawk-3:
fixes #1019


julesdesmit (approved):
LGTM

PR #1026, opened 2021-04-21
Add UNIX prefix to TPS util address in case we're using UNIX sockets

### LINKED ISSUE #940: Deploy transaction stress-testing utility on the DevNet
We would like to see the upper bound of processing capacity on a production-like environment. This work will most likely need to be coordinated with @pmoranga.

julesdesmit:
This PR also adds logging entries in the mempool which show the size of the mempool before and after processing blocks.

Fixes #940


autholykos (approved):
LGTM

PR #1027, opened 2021-04-21
Drop transactions if the mempool reaches max capacity

### LINKED ISSUE #1024: Fix potential OOM errors in the mempool
When running TPS stress tests on the devnet, it appeared that nodes crashed quite quickly from running out of RAM space. This is because the mempool is unbounded in how many items it allows to be stored.

A potential fix could be to store transactions to disk, taking the heavy load off the RAM, and allowing to accept a much higher limit of transactions before the mempool is filled. However, this is still up to discussion.

julesdesmit:
As discussed with goshawk-3, having a mempool grow
beyond its maximum capacity should be an indicator that
we should be looking at improving TPS. So for now, this
adjustment should be an adequate fix.

Fixes #1024

### DIFF
--- a/pkg/core/mempool/mempool.go
+++ b/pkg/core/mempool/mempool.go
@@ -161,6 +161,14 @@ func (m *Mempool) Run(ctx context.Context) {
 
 // ProcessTx handles a submitted tx from any source (rpcBus or eventBus).
 func (m *Mempool) ProcessTx(srcPeerID string, msg message.Message) ([]bytes.Buffer, error) {
+	maxSizeBytes := config.Get().Mempool.MaxSizeMB * 1000 * 1000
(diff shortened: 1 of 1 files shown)


autholykos (changes requested):
LGTM. Can merge after looking at why one of the test is failing (consensus-related. I don't think it has anything to do with this change)

julesdesmit:
@autholykos It appears that this is an intermittent error caused in the consensus testbed. The message unmarshaller complains about a malformed bn256 point, which makes me believe that this probably has to do with the message being passed around different nodes, as we've observed that verified signatures change their underlying representation. I will file an issue to tackle this later, but as it has not appeared on any production-like environment, I won't give it high priority for now.

Issue #819, opened 2020-12-21
Modularize networking-dependent components


autholykos:
See https://github.com/dusk-network/dusk-blockchain/pull/802#discussion_r546754391

julesdesmit:
This should include both the incoming as well as outgoing message processors. Ideally, when using Kadcast we should still just disseminate messages through the `Gossip` topic.

PR #1030, opened 2021-04-22
Reintegrate Kadcast to dusk-blockchain

### LINKED ISSUE #957: Stabilize test-harness TestMultipleBiddersProvisioners test with Kadcast enabled
Currently, Kadcast is capable of propagating topics.Block and topics.Tx. This issue is to make `TestMultipleBiddersProvisioners` runnable with Kadcast enabled.  Eventually, a CI test should be added to run `TestMultipleBiddersProvisioners` with Kadcast.

goshawk-3:
fixes #957 

Any test harness unit can be executed with Kadcast network if variable is set `DUSK_NETWORK_PROFILE=kadcast`.  An example make target: `test-harness-ci-kadcast`. Once #1033 is done, `test-harness-ci-kadcast` should be added to CI too.


cmd/dusk/server.go
```diff
@@ -84,15 +84,15 @@ func LaunchChain(ctx context.Context, cl *loop.Consensus, proxy transactions.Pro
 	return chainProcess, nil
 }
 
-func (s *Server) launchKadcastPeer() {
+func (s *Server) launchKadcastPeer(procesor *peer.MessageProcessor) {
```
julesdesmit, line 87: processor*

cmd/dusk/server.go
```diff
@@ -84,15 +84,15 @@ func LaunchChain(ctx context.Context, cl *loop.Consensus, proxy transactions.Pro
...
 	kcfg := cfg.Get().Kadcast
 
 	if !kcfg.Enabled {
 		log.Warn("Kadcast service is disabled")
 		return
 	}
 
-	kadPeer := kadcast.NewPeer(s.eventBus, s.gossip, nil, kcfg.Raptor)
+	kadPeer := kadcast.NewPeer(s.eventBus, s.gossip, nil, procesor, kcfg.Raptor)
```
julesdesmit, line 95: Ditto

pkg/core/chain/chain.go
```diff
@@ -64,8 +64,8 @@ type Loader interface {
 
 // Ledger is the Chain interface used in tests.
 type Ledger interface {
-	TryNextConsecutiveBlockInSync(block.Block) error
-	TryNextConsecutiveBlockOutSync(block.Block) error
+	TryNextConsecutiveBlockInSync(blk block.Block, kadcastHeight byte) error
+	TryNextConsecutiveBlockOutSync(blk block.Block, kadcastHeight byte) error
```
julesdesmit, line 68: What exactly is this kadcast height for? I am a little bit fuzzy on the paper at this point and I can't really figure out the use of this from the code.

julesdesmit (changes requested):
Aside from the current feedback, this is fine for now. There are some broader architectural things which I'm not a fan of (having both a `propagateBlock` and a `kadcastBlock` for instance, which can be easily abstracted over IMO), but these issues should be addressed when working on #819.

pkg/core/chain/chain.go
goshawk-3: `kadcastHeight`  (a.k.a `BCAST_HEIGHT` or `height` ) is kind of an identifier of a subtree in the binary routing tree (Kademlia). It should be propagated with each Kadcast wire message so that the receiver node knows which subtree  the message should re-propagated to.  The broadcast algorithm is well-illustrated at `figure-2` (https://eprint.iacr.org/2019/876.pdf).


NB. I don't like the need of pushing a transport-related field to the upper layers either. But this approach is stateless, lite and compliant with Kadcast protocol. Later on, we might consider a stateful solution where `kadcast Height` is known only by P2P layer.

goshawk-3:
> Aside from the current feedback, this is fine for now. There are some broader architectural things which I'm not a fan of (having both a `propagateBlock` and a `kadcastBlock` for instance, which can be easily abstracted over IMO), but these issues should be addressed when working on #819.

`propagateBlock` builds a `topics.Inv` message  with payload of a block hash whereas `kadcastBlock` builds a Kadcast message with payload of a full block plus it includes a header field. But if you find a better way to abstract these two, feel free to push it as a PR.

julesdesmit:
> propagateBlock builds a topics.Inv message with payload of a block hash whereas kadcastBlock builds a Kadcast message with payload of a full block plus it includes a header field. But if you find a better way to abstract these two, feel free to push it as a PR.

This can probably be abstracted over by simply broadcasting the block hash over the event bus, and allowing a component to take action based on the configuration (I envision two different ones, and depending on the config, a different one is hooked up to the event bus on startup). I'm currently not focused on the kadcast work, so I'll refrain from pushing work on it for now, but it may be good to keep in mind for work related to #819.

pkg/core/chain/chain.go
julesdesmit: I see. The reason I asked as well, is because I don't really see it mutated anywhere, and it seems that, whenever you need it, you just use the pre-configured initial kadcast height. Is this due to be changed? If so, may be worthwhile to log that in an issue.

goshawk-3: The kadcast height is decremented in `(w *Writer) broadcastPacket` on re-propagating a message. The Initial Kadcast height is only used when a node triggers a broadcast algorithm initially.

julesdesmit: Gotcha, makes sense.

julesdesmit (approved):
LGTM

PR #1031, opened 2021-04-22
Refactor server and voucher seeder to allow for fetching of addresses at any time

### LINKED ISSUE #253: Add service flags, and distinguish between light and full node peers

### LINKED ISSUE #1021: Allow to contact the voucher seeder mid-process in case too many connections are dropped
As seen in #1020, an OS-related error (such as a broken pipe) can cause unforeseen disconnects for many different peers in a node. When the panic-causing code is removed it would also be ideal for the node to have a defensive response, which is to re-establish lost connections. Since we can not know for sure if the node on the other end is the cause of the trouble, it would be wise to instead just contact the voucher seeder again, and obtain a list of active nodes, to assimilate itself back into the network.

julesdesmit:
Fixes #1021, fixes #253


cmd/dusk/server.go
```diff
@@ -244,16 +243,25 @@ func Setup() *Server {
 	// Creating the peer factory
 	readerFactory := peer.NewReaderFactory(processor)
 
+	// Create the listener and contact the voucher seeder
+	connector := peer.NewConnector(cfg.Get().Network.Port)
+
+	seeders := cfg.Get().Network.Seeder.Addresses
+	if len(seeders) != 0 {
+		if err := connector.Connect(seeders[0]); err != nil {
```
devtomnl, line 251: if this is an array, would not make sense to iterate over it and try with more than the position zero ?

pkg/p2p/peer/connector.go
```diff
@@ -0,0 +1,161 @@
...
+		registry:      make(map[string]struct{}),
+		services:      services,
+	}
+
+	go func(c *Connector) {
+		for {
+			conn, err := c.l.Accept()
+			if err != nil {
+				log.WithField("process", "connection manager").
+					WithError(err).
```
devtomnl, line 64: would you agree that for debugging proposes would make sense to also add more info such as addrPort ?

goshawk-3 (changes requested):
Running `make test-harness-ci`  where network size is 3 nodes (which is the default CI action) passes successfully. But when running test harness with  network size higher than 3 nodes (e.g `make test-harness-alive`)  it seems that a part of the network is stuck at height 0. I suppose that the nodes after the third one did not manage to join the network somehow.

@julesdesmit 


```
NODE:9500:
{"data":{"blocks":[{"header":{"height":14}}]}}
NODE:9501:
{"data":{"blocks":[{"header":{"height":14}}]}}
NODE:9502:
{"data":{"blocks":[{"header":{"height":14}}]}}
NODE:9503:
{"data":{"blocks":[{"header":{"height":0}}]}}
NODE:9504:
{"data":{"blocks":[{"header":{"height":0}}]}}
NODE:9505:
{"data":{"blocks":[{"header":{"height":0}}]}}
NODE:9506:
{"data":{"blocks":[{"header":{"height":0}}]}}
NODE:9507:
{"data":{"blocks":[{"header":{"height":0}}]}}
NODE:9508:
{"data":{"blocks":[{"header":{"height":0}}]}}
```

```
 level=info msg="Network at round [22] driven by nodes [ 0, 1, 2,]" process=monitor
 level=info msg="Network at round [0] driven by nodes [ 3, 4, 5, 6, 7, 8,]" process=monitor
 level=error msg="network inconsistency detected"
```

julesdesmit:
@goshawk-3 It appears that the challenge message is sent too early in some cases, causing the node not to be able read it (after 60 seconds, it reports an i/o timeout). I have achieved a consistently synchronized network many times over by introducing a 3 second sleep, before sending the challenge over to the node. A fix will be pushed shortly.

pkg/p2p/peer/connector.go
```diff
@@ -0,0 +1,198 @@
...
+		eventBus:      eb,
+		gossip:        gossip,
+		readerFactory: NewReaderFactory(processor),
+		l:             listener,
+		registry:      make(map[string]struct{}),
+		services:      services,
+		connectFunc:   connectFunc,
+	}
+
+	processor.Register(topics.Addr, c.ProcessNewAddress)
```
goshawk-3, line 67: What if an adversary node is sending `invalid` peer addresses via message `topics.Addr`, this again could allow a Sybil attack.

pkg/p2p/peer/connector.go
```diff
@@ -0,0 +1,198 @@
...
+// Close the listener.
+func (c *Connector) Close() error {
+	return c.l.Close()
+}
+
+// ProcessNewAddress will handle a new Addr message from the network.
+// Satisfies the peer.ProcessorFunc interface.
+func (c *Connector) ProcessNewAddress(srcPeerID string, m message.Message) ([]bytes.Buffer, error) {
+	a := m.Payload().(message.Addr)
+	return nil, c.Connect(a.NetAddr)
```
goshawk-3, line 95: An OS usually has fd max limit, if a node is flooded with `topics.Addr`, this limit could be reached and the goroutine would panic. Shall we limit here the number of dialing a connection?

cmd/voucher/challenger/challenger.go
```diff
@@ -0,0 +1,99 @@
...
+
+	buf := bytes.NewBuffer(challenge)
+	if err := topics.Prepend(buf, topics.Challenge); err != nil {
+		log.Panic(err)
+	}
+
+	ch <- *buf
+
+	// Enter the node into the store, for future reference.
+	c.nodes.Add(r.Addr(), challenge)
```
goshawk-3, line 51: A kind of a race condition (very rare situation I believe), the voucher might receive `topics.Response` message before adding the peer to the registry thus the peer would be blacklisted. Shall we first add the peer and then send the wire message?

harness/engine/profiles.go
```diff
@@ -64,7 +66,7 @@ func Profile1(index int, node *DuskNode, walletPath string) {
 	viper.Set("timeout.timeoutkeepalivetime", 30)
 
 	viper.Set("logger.output", node.Dir+"/dusk")
-	viper.Set("logger.level", "info")
+	viper.Set("logger.level", "trace")
```
goshawk-3, line 69: Maybe a typo? we should have `info` here, right?

pkg/p2p/peer/connector.go
```diff
@@ -0,0 +1,198 @@
...
+		return err
+	}
+
+	go c.proposeConnection(conn)
+	return nil
+}
+
+// Dial dials up a connection, given its address string.
+func (c *Connector) Dial(addr string) (net.Conn, error) {
+	dialTimeout := 1 * time.Second
```
goshawk-3, line 112: Maybe this shouldn't be hard coded.

cmd/dusk/server.go
```diff
@@ -244,16 +239,26 @@ func Setup() *Server {
 	// Creating the peer factory
 	readerFactory := peer.NewReaderFactory(processor)
 
+	// Create the listener and contact the voucher seeder
+	gossip := protocol.NewGossip(protocol.TestNet)
+	connector := peer.NewConnector(eventBus, gossip, cfg.Get().Network.Port, processor, protocol.ServiceFlag(cfg.Get().Network.ServiceFlag), peer.Create)
+
+	seeders := cfg.Get().Network.Seeder.Addresses
+	for _, seeder := range seeders {
+		if err = connector.Connect(seeder); err != nil {
```
goshawk-3, line 248: If a node fails to connect to all vouchers here we should at least panic for now. Later on, we can implement a reattempt-to-connect procedure.

pkg/p2p/peer/connector.go
goshawk-3: In addition, `1 sec` might not be enough in a busy network with high network latency. Most probably,  the test-harness node that failed to connect to the Voucher with `i/o timeout` error recently was due to this limit.

goshawk-3 (changes requested):

cmd/voucher/challenger/challenger.go
julesdesmit: It would be impossible, because the read loop isn't instantiated until the node is added to the list. The message might have arrived, but it won't be read yet.

harness/engine/profiles.go
julesdesmit: I swapped this in order to debug some things and accidentally committed it. Will revert.

cmd/dusk/server.go
julesdesmit: Agreed. Will extrapolate into an issue

cmd/voucher/challenger/challenger.go
goshawk-3: Makes sense!

pkg/p2p/peer/connector.go
```diff
@@ -0,0 +1,198 @@
...
+	// Ensure we are still above the minimum connections threshold.
+	if len(c.registry) < config.Get().Network.MinimumConnections {
+		buf := new(bytes.Buffer)
+		if err := topics.Prepend(buf, topics.GetAddrs); err != nil {
+			log.WithField("process", "peer connector").
+				WithError(err).
+				Panic("could not create topic buffer")
+		}
+
+		c.eventBus.Publish(topics.Gossip, message.New(topics.GetAddrs, *buf))
```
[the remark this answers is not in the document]
julesdesmit: As discussed, this should be handled by the implementation of a reputation system.

pkg/p2p/peer/connector.go
julesdesmit: Same as above.

goshawk-3 (approved):
LGTM

PR #1038, opened 2021-05-05
attempt to debug & fix #1035 

### LINKED ISSUE #1035: Investigate the cause of timeouts in the automatons
When initiating the automatons in the devnet, these messages pop up on all nodes:

```
{"error":"timeout-ed request","level":"error","msg":"could not send bid tx","process":"BidAutomaton","time":"2021-04-30T14:00:48Z"}
{"error":"timeout-ed request","level":"error","msg":"could not send stake tx","process":"StakeAutomaton","time":"2021-04-30T14:00:48Z"}
```

After increasing the timeout to 10 seconds, this message still appears. The cause is evidently not from expensive computation, so the real cause must be identified and addressed.

devtomnl:
In order to try to address issue reported on #1035 , I decided to see if increasing the timeouts would be of any help.

* add option to configure automate job stake and bid timeout request calls to provisioner and blockgenerator in order to figure out if this helps to address the issue

Please let me know where else could I look.

### DIFF
--- a/cmd/utils/grpcclient/automatonservice.go
+++ b/cmd/utils/grpcclient/automatonservice.go
@@ -15,7 +15,7 @@ import (
 )
 
 // AutomateStakesAndBids will enable the use of the stake and bid automaton in a node.
-func AutomateStakesAndBids(address string) error {
+func AutomateStakesAndBids(address string, sendStakeTimeout, sendBidTimeout int) error {
 	// Add UNIX prefix in case we're using unix sockets.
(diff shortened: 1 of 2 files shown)


julesdesmit (approved):
This is worth a try, and it's good to add this configurability. I think that the best way forward, however, is to profile the execution and see why it takes so long. So let's merge this, but let's keep in mind that we should take a proper look at what's causing the slowdown.

PR #1039, opened 2021-05-05
Use blake2b for computing chunk IDs instead of Sha3. #207

devtomnl:
* replace sha3 by blake2b on kadcast helpers and peerinfo


julesdesmit (approved):
LGTM

PR #1040, opened 2021-05-11
Introduce point-to-point messaging in Kadcast

### LINKED ISSUE #1033: Allow a peer in Kadcast network to sync up properly
**Problem**
Any node in Kadcast network falling behind (getting into `OutSync` state)  fails to download any blocks.  Normally, in Gossip network a node is capable to request missing blocks via `topics.GetBlocks` relying on the persistent TCP connection with its peer. More particularly, the `Peer.writeloop` utilizes `respChan` to write to the already established TCP connection. 

**Solution**
When a Kadcast node is in `OutSync` state it should establish a TCP connection with the initiator Peer, request missing blocks via `topics.GetBlocks` and close the connection.  In this way, we could reuse fully or partly current gossip-related implementation of Synchronizer.

goshawk-3:
fixes #1033 

As a result from this PR, new github action has been added - `Continuous Integration / test-harness-kadcast`. It's goal is to continuously retest dusk network/consensus with Kadcast+RCUDP enabled.

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -47,3 +47,14 @@ jobs:
       uses: actions/checkout@v2
     - name: Test Harness
       run: make test-harness-ci
+  test-harness-kadcast:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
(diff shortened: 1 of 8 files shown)


autholykos (approved):
LGTM. Good job

PR #1042, opened 2021-05-18
Integrate topics.Score propagation into Kadcast network

### LINKED ISSUE #1041: Integrate topics.Score message broadcast into Kadcast network
As per recent refactoring, Candidate block is now propagated as a payload item of `topics.Score` wire message. In addition, on recent TPS benchmark tests it turns out that the size of Candidate block has significant impact over average block time, respectively network TPS (see also #1019). Based on these findings, we need to integrate topics.Score message into Kadcast before comparing its Kadcast TPS capabilities to Gossip.

### LINKED ISSUE #1048: Alter selection phase so it broadcasts bestScore only
It  was found that a node is publishing  same` topics.Score` message twice when the score turns to be the highest one. This consensus message brings Candidate block which respectively has impact over consensus average time. An optimization here could be to publish only the best score as per the protocol specification.

This should be verified by test-harness reaching 10K rounds as a regression test.

goshawk-3:
fixes #1041 fixes #1048

### DIFF
--- a/pkg/core/consensus/selection/step.go
+++ b/pkg/core/consensus/selection/step.go
@@ -102,14 +102,7 @@ func (p *Phase) generateCandidate(ctx context.Context, round consensus.RoundUpda
 	log.
 		WithField("step", step).
 		WithField("round", round.Round).
-		Debugln("sending score")
-
-	// create the message
-	msg := message.New(topics.Score, *scr)
-
-	if err := p.Republish(msg, []byte{config.KadcastInitialHeight}); err != nil {
-		lg.WithError(err).Errorln("topics.Score publishing failed")
-	}
+		Debugln("sending score internally")
 
 	// communicate our own score to the selection
 	internalScoreChan <- message.New(topics.Score, *scr)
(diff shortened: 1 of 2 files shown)


pkg/core/consensus/selection/step.go
```diff
@@ -142,17 +142,19 @@ func (p *Phase) Run(parentCtx context.Context, queue *consensus.Queue, evChan ch
...
 		}
 	}
 
 	for {
 		select {
 		case internalScoreResult := <-internalScoreChan:
-			p.collectScore(ctx, internalScoreResult.Payload().(message.Score))
+			// disable topics.Score network republish here as it's been already published in generateCandidate.
+			const republish = false
+			p.collectScore(ctx, internalScoreResult.Payload().(message.Score), internalScoreResult.Header(), republish)
```
julesdesmit, line 154: As discussed before on the standup, I think it would be cleaner to remove the republishing in the candidate generator, and then avoid the boolean flag since we can always republish in `collectScore`. As checked with Toghrul, it's legal in the protocol to avoid republishing one's own `Score` message in case a better one was already processed.

julesdesmit (changes requested):

pkg/core/consensus/selection/step.go
goshawk-3: Indeed. I should probably remove republish flag from this PR and handle the "optimization" with a separate issue/PR. This is because a such protocol-related change should pass 10K-round test to ensure no regressions occur. I still have a hunch that we had similar idea implemented time ago and that had negative impact on the consensus average time. Need to test this more.

goshawk-3: @julesdesmit Feel free to review description of #1048

julesdesmit (approved):
LGTM

PR #1050, opened 2021-05-19
Add automaton to devnet.sh and fix sendbid tx

devtomnl:
How to test:
Terminal 1: ./devnet.sh -q 3 -r true -b true -d true -a true
Terminal 2: make build && ./bin/utils metrics
Terminal 3: watch "curl http://localhost:9099/metrics

Check logs and make sure there are no errs such as: `rpc error: code = Unauthenticated desc = token not provided`


julesdesmit (approved):
LGTM

PR #1052, opened 2021-05-20
#1050 - add response to automaton requests

devtomnl:
in order to stop having timeout errs in automate bid and stake requests, I've added responses to the requests and this seems to clear the logs.

How to test:
run devnet with automate true
`./devnet.sh -q 3 -r true -b true -d true -a true`

Observe there are no timeout errs.


julesdesmit (approved):
LGTM

PR #1057, opened 2021-05-27
Optimize broadcast implementation in kadcast

### LINKED ISSUE #1043: Measure Kadcast network transactions-per-second
As per the benchmark tests done for Gossip (see also #1019), we should measure and make reports for Kadcast network ran in different environments settings. This might require  integrating Kadcast into DevNet env as well.

Pending: List here scenarios to be used for benchmark tests.

goshawk-3:
fixes #1043

### DIFF
--- a/pkg/p2p/kadcast/maintainer.go
+++ b/pkg/p2p/kadcast/maintainer.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/kadcast/encoding"
+	"github.com/sirupsen/logrus"
 	logger "github.com/sirupsen/logrus"
 )
 
@@ -248,8 +249,10 @@ func (m *Maintainer) send(raddr net.UDPAddr, payload []byte) {
 
 	_ = conn.SetDeadline(time.Now().Add(10 * time.Second))
 
-	log.WithField("src", conn.LocalAddr().String()).
-		WithField("dest", raddr.String()).Traceln("Sending udp")
+	if log.Logger.GetLevel() == logrus.TraceLevel {
+		log.WithField("src", conn.LocalAddr().String()).
+			WithField("dest", raddr.String()).Traceln("Sending udp")
+	}
 
 	// Simple write
 	_, err = conn.Write(payload)
--- a/pkg/p2p/kadcast/utils.go
+++ b/pkg/p2p/kadcast/utils.go
@@ -271,21 +271,6 @@ func getRandDelegatesByShuffle(beta uint8, in []encoding.PeerInfo) []encoding.Pe
 	return out
 }
 
-// getNoRandDelegates selects n in a row items from `in` without any rand function.
-// Used for benchmarking.
-func getNoRandDelegates(beta uint8, in []encoding.PeerInfo) []encoding.PeerInfo {
-	out := make([]encoding.PeerInfo, 0, beta)
-
-	for i := 0; i < int(beta); i++ {
-		if i >= len(in) {
-			break
-		}
-
-		out = append(out, in[i])
-	}
-	return out
-}
-
 func makeHeader(t byte, rt *RoutingTable) encoding.Header {
 	return encoding.Header{
 		MsgType:         t,
--- a/pkg/p2p/kadcast/utils.go
+++ b/pkg/p2p/kadcast/utils.go
@@ -14,6 +14,7 @@ import (
 	"io"
 	"math/big"
 	"math/bits"
+	mathrand "math/rand"
 	"net"
 	"strconv"
 	"time"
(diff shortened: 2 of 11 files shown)


pkg/p2p/kadcast/maintainer.go
```diff
@@ -248,8 +249,10 @@ func (m *Maintainer) send(raddr net.UDPAddr, payload []byte) {
 
 	_ = conn.SetDeadline(time.Now().Add(10 * time.Second))
 
-	log.WithField("src", conn.LocalAddr().String()).
-		WithField("dest", raddr.String()).Traceln("Sending udp")
+	if log.Logger.GetLevel() == logrus.TraceLevel {
```
julesdesmit, line 252: This isn't necessary, is it? The call will just be ignored if we don't set the logger to capture `trace`.

pkg/p2p/kadcast/utils.go
```diff
@@ -247,7 +248,42 @@ func generateRandomDelegates(beta uint8, in []encoding.PeerInfo, out *[]encoding
...
+
+		out = append(out, in[i])
+	}
+
+	return out
+}
+
+// getNoRandDelegates selects n in a row items from `in` without any rand function.
+// Used for benchmarking.
+func getNoRandDelegates(beta uint8, in []encoding.PeerInfo) []encoding.PeerInfo {
```
julesdesmit, line 276: Probably amend that comment, since you also use it in the actual code - I was under the impression this was for testing only.

pkg/p2p/kadcast/writer.go
```diff
@@ -115,51 +116,79 @@ func (w *Writer) WriteToPoint(m message.Message) error {
...
-
-	if height > byte(len(router.tree.buckets)) || height == 0 {
+func (w *Writer) broadcastPacket(maxHeight byte, payload []byte) {
+	if maxHeight > byte(len(w.router.tree.buckets)) || maxHeight == 0 {
 		return
 	}
 
-	log.WithField("l_addr", myPeer.String()).WithField("height", height).
-		Traceln("Run broadcasting algorithm")
+	if log.Logger.GetLevel() == logrus.TraceLevel {
```
julesdesmit, line 144: Same here, this `if` seems unnecessary

pkg/p2p/kadcast/writer.go
```diff
@@ -115,51 +116,79 @@ func (w *Writer) WriteToPoint(m message.Message) error {
...
 // BroadcastPacket sends a `CHUNKS` message across the network
 // following the Kadcast broadcasting rules with the specified height.
-func (w *Writer) broadcastPacket(height byte, payload []byte) {
-	router := w.router
-	myPeer := router.LpeerInfo
-
-	if height > byte(len(router.tree.buckets)) || height == 0 {
+func (w *Writer) broadcastPacket(maxHeight byte, payload []byte) {
+	if maxHeight > byte(len(w.router.tree.buckets)) || maxHeight == 0 {
 		return
```
julesdesmit, line 141: Would it be useful to return an error here, to let the caller know that we aren't broadcasting a packet?

pkg/p2p/kadcast/writer.go
```diff
@@ -185,62 +214,48 @@ func (w *Writer) fetchDelegates(H byte) []encoding.PeerInfo {
...
 			continue
 		}
 
-		log.WithField("l_addr", localPeer.String()).
-			WithField("r_addr", destPeer.String()).
-			WithField("height", H).
-			WithField("raptor", w.raptorCodeEnabled).
-			WithField("len", buf.Len()).
-			Trace("Sending message")
+		if log.Logger.GetLevel() == logrus.TraceLevel {
```
julesdesmit, line 239: Again, seems redundant.

julesdesmit (changes requested):

pkg/p2p/kadcast/maintainer.go
goshawk-3: It is needed due to the `WithField` calls prior to Traceln. It was spotted by the cpu profiler.

pkg/p2p/kadcast/utils.go
goshawk-3: This should be temporary. Before opening the PR for review, I'll probably remove it.

pkg/p2p/kadcast/writer.go
goshawk-3: Same as above ^

pkg/p2p/kadcast/writer.go
```diff
@@ -115,51 +116,79 @@ func (w *Writer) WriteToPoint(m message.Message) error {
...
-		Traceln("Run broadcasting algorithm")
+	if log.Logger.GetLevel() == logrus.TraceLevel {
+		log.WithField("l_addr", w.router.LpeerInfo.String()).WithField("max_height", maxHeight).
+			Traceln("broadcasting procedure")
+	}
+
+	// Marshal message data
+	packet, err := w.marshalBroadcastPacket(0, payload)
+	if err != nil {
+		log.WithError(err).Warn("marshal broadcast packet failed")
```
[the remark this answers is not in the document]
goshawk-3: Good catch!

pkg/p2p/kadcast/writer.go
goshawk-3: That actually makes me reconsider error reporting in general. Addressed it with latest commit.

julesdesmit (approved):
LGTM

PR #1063, opened 2021-05-31
Clean up timer resources properly when using RPCBus

### LINKED ISSUE #1062: Clean up timer resources in RPCBus
Memory profiling revealed to us that heavy usage of the RPCBus causes lots of lingering timers, due to the use of `time.After`.

julesdesmit:
Fixes #1062

### DIFF
--- a/pkg/util/nativeutils/rpcbus/rpcbus.go
+++ b/pkg/util/nativeutils/rpcbus/rpcbus.go
@@ -130,17 +130,29 @@ func (bus *RPCBus) Call(t topics.Topic, req Request, timeOut time.Duration) (int
 }
 
 func (bus *RPCBus) callTimeout(reqChan chan<- Request, req Request, timeOut time.Duration) (interface{}, error) {
+	timer := time.NewTimer(timeOut)
+
 	select {
 	case reqChan <- req:
-	case <-time.After(timeOut):
+	case <-timer.C:
 		return bytes.Buffer{}, ErrRequestTimeout
 	}
+
+	if !timer.Stop() {
+		<-timer.C
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/util/nativeutils/rpcbus/rpcbus.go
```diff
@@ -130,19 +130,31 @@ func (bus *RPCBus) Call(t topics.Topic, req Request, timeOut time.Duration) (int
...
+
 	select {
 	case reqChan <- req:
-	case <-time.After(timeOut):
+	case <-timer.C:
 		return bytes.Buffer{}, ErrRequestTimeout
 	}
 
+	if !timer.Stop() {
+		<-timer.C
```
goshawk-3, line 142: Should we drain this chan in a separate goroutine so we don't block `callTimeout`

julesdesmit:
I don’t think that’s necessary, at most we are draining a single element.
If it becomes a performance bottleneck, we can always optimize later.

Op ma 31 mei 2021 om 19:20 schreef goshawk-3 ***@***.***>

pkg/util/nativeutils/rpcbus/rpcbus.go
julesdesmit: I don’t think that’s necessary, at most we are draining a single element.
If it becomes a performance bottleneck, we can always optimize later.

goshawk-3: To be honest, I couldn't find where in `timerproc` or `deltimerLocked` the timer callback (the one that writes to timer.C) is called but let's rely on the example from documentation.

goshawk-3 (approved):

PR #1067, opened 2021-06-02
Integrate Height byte field into UDP packet

### LINKED ISSUE #1058: Implement K-Height transmission as a UDP package data field
Latest implementation transmits Height byte (K-Height/`BCAST_HEIGHT`) as part of Broadcast message header. For making this possible, broadcast procedure must rebuild Raptor UDP blocks each time a message is sent to another sub-tree.

As per suggestion from specification, K-Height byte must be transmitted with each UDP (chunk) packet thus avoiding the need of Raptor Block recompiling each time.

goshawk-3:
fixes #1058

### DIFF
--- a/pkg/util/nativeutils/rcudp/config.go
+++ b/pkg/util/nativeutils/rcudp/config.go
@@ -26,7 +26,7 @@ const (
 	// Encoder configs.
 
 	// BlockSize - max length of an encoding symbol that can fit into a single wire packet.
-	BlockSize = maxUDPLength - 8 - 2 - 2 - 4 - 4
+	BlockSize = maxUDPLength - 8 - 2 - 2 - 4 - 4 - 8
 
 	// symbolAlignmentSize = Al is the size of each symbol in the source message in bytes.
 	// Usually 4. This is the XOR granularity in bytes. On 32-byte machines 4-byte XORs.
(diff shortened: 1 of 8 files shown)


pkg/util/nativeutils/rcudp/config.go
```diff
@@ -26,7 +26,7 @@ const (
 	// Encoder configs.
 
 	// BlockSize - max length of an encoding symbol that can fit into a single wire packet.
-	BlockSize = maxUDPLength - 8 - 2 - 2 - 4 - 4
+	BlockSize = maxUDPLength - 8 - 2 - 2 - 4 - 4 - 8
```
julesdesmit, line 29: It seems like this is knocking off all the packet fields from the max UDP frame size. In this case, shouldn't you use `- 1` for the `bcastHeightSize` field instead?

autholykos: Seconded. I doubt that we'd need more than one byte for the block size height

autholykos (changes requested):

pkg/util/nativeutils/rcudp/config.go
goshawk-3: No, it cannot be 1 byte.  As per RFC 5053 `- The symbol size, T, in bytes, which MUST be a multiple of Al` .

autholykos (approved):

PR #1071, opened 2021-06-04
Remove respChan and writeLoop, use response callback directly

### LINKED ISSUE #1070: Fix bottleneck in the `respChan`

julesdesmit:
This alleviates the backpressure observed during
times of high load, and simplifies the Peer structure.

Fixes #1070


julesdesmit:
A run on devnet did not actually relieve the memory leak. A ring buffer may be a more suitable solution, since we should manage the amount of running goroutines.

PR #1072, opened 2021-06-04
Use ringbuffer directly in favor of respChan

### LINKED ISSUE #1070: Fix bottleneck in the `respChan`

julesdesmit:
This alleviates the backpressure observed during
times of high load, and simplifies the Peer structure.

Fixes #1070

### DIFF
--- a/pkg/p2p/peer/peer.go
+++ b/pkg/p2p/peer/peer.go
@@ -215,8 +215,9 @@ func Create(ctx context.Context, reader *Reader, writer *Writer) {
 	defer cancel()
 
 	g := &GossipConnector{writer.Connection}
-	ringBuf, l := eventbus.NewStreamListener(g)
+	l := eventbus.NewStreamListener(g)
 	writer.gossipID = writer.subscriber.Subscribe(topics.Gossip, l)
+	ringBuf := ring.NewBuffer(1000)
 
 	// On each new connection the node sends topics.Mempool to retrieve mempool
 	// txs from the new peer
@@ -225,6 +226,8 @@ func Create(ctx context.Context, reader *Reader, writer *Writer) {
 		logrus.WithField("process", "peer").
 			Errorln("could not send mempool message to peer")
 	}
+
+	_ = ring.NewConsumer(ringBuf, eventbus.Consume, g)
 
 	reader.ReadLoop(pCtx, ringBuf)
 	writer.onDisconnect()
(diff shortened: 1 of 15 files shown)


pkg/p2p/peer/peer.go
```diff
@@ -329,10 +278,10 @@ func (p *Reader) ReadLoop(ctx context.Context) {
 	// 		log.Errorf("Peer %s failed with critical issue: %v", p.RemoteAddr(), r)
 	// 	}
 	// }()
-	p.readLoop(ctx)
+	p.readLoop(ctx, ringBuf)
 }
 
-func (p *Reader) readLoop(ctx context.Context) {
+func (p *Reader) readLoop(ctx context.Context, ringBuf *ring.Buffer) {
```
autholykos, line 284: using a private `readLoop` makes no sense here. Please incorporate this function in the exported one

pkg/p2p/peer/processor.go
```diff
@@ -46,7 +47,7 @@ func (m *MessageProcessor) Register(topic topics.Topic, fn ProcessorFunc) {
 
 // Collect a message from the network. The message is unmarshaled and passed down
 // to the processing function.
-func (m *MessageProcessor) Collect(srcPeerID string, packet []byte, respChan chan<- bytes.Buffer, services protocol.ServiceFlag, header []byte) ([]bytes.Buffer, error) {
+func (m *MessageProcessor) Collect(srcPeerID string, packet []byte, ringBuf *ring.Buffer, services protocol.ServiceFlag, header []byte) ([]bytes.Buffer, error) {
```
autholykos, line 50: rename `ringBuffer` into `respRingBuf` (or `responsePipeline` if you wanna be more generic)

autholykos (changes requested):
I like the idea of pushing messages to a `ringBuffer` rather than using a channel. However, I am afraid that reusing the same `ringBuffer` created in the `eventbus.StreamListener` ends up breaking the encapsulation of the message passing system used by the other components. I would instantiate a second dedicated `ringBuffer` explicitly in the `peer.Create`, with a preloaded `Mempool` topic, to be passed and used exclusively by to the `Reader` for handling responses from the peer, rather than overloading the `StreamListener`'s one (which is intended for many-to-one communication of outbound messages). This is to be preferred especially because the `peer.Writer` seems to never use the `ringBuffer` explicitly (except in the `peer.Create` for the `Mempool` topic)

autholykos (approved):
LGTM

PR #1077, opened 2021-06-09
Reduce frequency at which the StormDB is called

### LINKED ISSUE #1076: Evaluate inefficiencies in use of StormDB
During TPS stress testing, it was observed that goroutine bloat was caused by an excessive amount of goroutines attempting to update StormDB with a peer count. The whole repo should be assessed and it needs to be made sure that this can't happen.

julesdesmit:
Fixes #1076

### DIFF
--- a/pkg/p2p/peer/connector.go
+++ b/pkg/p2p/peer/connector.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/config"
+	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/capi"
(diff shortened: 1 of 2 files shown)


goshawk-3 (approved):
LGTM

devtomnl:
I updated dusk.toml on my local devnet.sh setup to test it with:

```
[api]
  enabled=true
```

Then after starting it, I ran the following cmds:

1. `watch "curl -s http://localhost:9490/p2p/count"`  
2.  `watch curl "http://localhost:9490/p2p/logs?type=Writer"`
3.  `watch curl "http://localhost:9490/p2p/logs?type=Reader"`

And all outputs seems to be as before, so it seems to work well. Cheers.

devtomnl (approved):

PR #1080, opened 2021-06-09
Eliminate separation between block generation and provisioning

### LINKED ISSUE #1078: Eliminate separation between block generation and provisioning
This entails the following changes:
- The block generator is now selected through the same deterministic sortition algorithm as the provisioners are
- No more proof-of-blind-bid generation and verification
- No more fixed-time selection step. One block is created, propagated, and immediately acted upon
- No more maintenance of a bid merkle tree
- Consensus can become dramatically faster due to reduced complexity
- Score message can be removed entirely
- More TBD

julesdesmit:
Fixes #1078


pkg/p2p/wire/message/score.go
```diff
@@ -8,147 +8,69 @@ package message
...
-		Score    []byte
-		Proof    []byte
-		Identity []byte
-		Seed     []byte
-	}
-
 	// Score extends the ScoreProposal with additional fields related to the
 	// candidate it pairs up with. The Score is supposed to be immutable once
 	// created and it gets forwarded to the other nodes.
 	Score struct {
```
autholykos, line 25: I guess we can rename this in a separate PR

autholykos (commented):
Have a few comments. However, given the size of the PR. It might be better to tackle them in a separate issue. Up to you. Depending on your reply I can approve or request changes

PR #1081, opened 2021-06-09
Fix devnet.sh #1079

ivensfernando:
How to test:
Terminal 1: `./devnet.sh -q 3 -r true -b true -d true -a true`
Terminal 2: `make build && ./bin/utils metrics`
Terminal 3: `watch "curl http://localhost:9099/metrics"`
Terminal 4: `watch "ps -ef |grep dusk"`

Check logs and make sure there are no errs and that there are four instances of Dusk, four instances of mockrusk, one instance of mock and one instance of voucher.


devtomnl:
LGTM

PR #1096, opened 2021-06-15
Remove APK and step from Agreement message

### LINKED ISSUE #271: Investigate removing the `Step` header attribute from the `StepVotes` within the firststep reducer
see [this comment](https://github.com/dusk-network/dusk-blockchain/pull/262#discussion_r373571357) for more info

### LINKED ISSUE #1085: Find cause of 'malformed point' error and fix it
In the devnet, using the new non-segregated consensus setup, the nodes occasionally run into `malformed point` errors, causing consensus slowdowns. Because of the error message, we can infer that it happens during unmarshalling. Since there are no consensus messages which do compression/decompression during marshalling except for the Agreement, this should be the target for investigation.

julesdesmit:
These items were not used at all, the step wasn't even unmarshaled.
This should get rid of our malformed point error, as it originated
in unmarshaling the APK, which now no longer exists on that message.

Fixes #271, fixes #1085

### DIFF
--- a/pkg/p2p/peer/processor.go
+++ b/pkg/p2p/peer/processor.go
@@ -49,10 +49,11 @@ func (m *MessageProcessor) Register(topic topics.Topic, fn ProcessorFunc) {
 // to the processing function.
 func (m *MessageProcessor) Collect(srcPeerID string, packet []byte, respRingBuf *ring.Buffer, services protocol.ServiceFlag, header []byte) ([]bytes.Buffer, error) {
 	b := bytes.NewBuffer(packet)
+	topic := topics.Topic(b.Bytes()[0])
 
 	msg, err := message.Unmarshal(b)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%s - topic: %s", err, topic)
 	}
 
 	if header != nil {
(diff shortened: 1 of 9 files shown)


pkg/p2p/peer/processor.go
```diff
@@ -49,10 +49,11 @@ func (m *MessageProcessor) Register(topic topics.Topic, fn ProcessorFunc) {
 // to the processing function.
 func (m *MessageProcessor) Collect(srcPeerID string, packet []byte, respRingBuf *ring.Buffer, services protocol.ServiceFlag, header []byte) ([]bytes.Buffer, error) {
 	b := bytes.NewBuffer(packet)
+	topic := topics.Topic(b.Bytes()[0])
```
goshawk-3, line 52: Just in case, let's check here if `b.Len()` or `len(packet)` is not 0

goshawk-3 (changes requested):
Would you confirm that we've got stable stressnet for at least 10000 rounds after introducing this fix? In other words, how can we ensure #1085 is completely fixed?

autholykos (approved):
It feels indeed like serializing the APK was just redundant. I can't really remember why we implemented it that way, but from reading the code, the only modification applied was removing the APK serialization and as a field of the `StepVotes` (and the `Step`, which it is clearly redundant). These changes LGTM, although it is true that we should stage this  carefully in devnet

PR #1111, opened 2021-07-02
Improve consensus implementation traceability 

### LINKED ISSUE #1103: Improve consensus internal state traceability

goshawk-3:
fixes #1103

### DIFF
--- a/pkg/util/ruskmock/server.go
+++ b/pkg/util/ruskmock/server.go
@@ -31,6 +31,7 @@ import (
 var log = logrus.WithField("process", "mock rusk server")
 
 const stateTransitionDelay = 1 * time.Second
+const stakeGracePeriod = 2
 
 // Server is a stand-in Rusk server, which can be used during any kind of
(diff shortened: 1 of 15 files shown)


pkg/util/ruskmock/server.go
```diff
@@ -30,7 +30,10 @@ import (
 
 var log = logrus.WithField("process", "mock rusk server")
 
-const stateTransitionDelay = 1 * time.Second
+const (
+	stateTransitionDelay = 1 * time.Second
+	stakeGracePeriod     = 2
```
julesdesmit, line 35: To stay in line with the consensus rules, we should probably keep this at `1000`. I guess you put it to `2` for testing purposes?

goshawk-3: Until we address all stability and performance issues in consensus, we should allow `stakeGracePeriod=2` in mocking service for both test-harness and devnet testing.

julesdesmit (approved):
LGTM

PR #1120, opened 2021-07-12
Limit max amount of messages in consensus queue

### LINKED ISSUE #1117: Implement a max size for the consensus queue
This should prevent OOM errors as we've seen them on the devnet.

julesdesmit:
Fixes #1117

### DIFF
--- a/pkg/core/consensus/agreement/step.go
+++ b/pkg/core/consensus/agreement/step.go
@@ -105,7 +105,10 @@ func (s *Loop) shouldCollectNow(a message.Message, round uint64, queue *consensu
 		return false
 	}
 
-	if hdr.Round > round {
+	// Only store events up to 10 rounds into the future
+	// XXX: According to protocol specs, we should abandon consensus
+	// if we notice valid messages from far into the future.
+	if hdr.Round > round && hdr.Round-round < 10 {
 		lg.
 			WithFields(log.Fields{
 				"topic":             "Agreement",
--- a/pkg/core/consensus/queue.go
+++ b/pkg/core/consensus/queue.go
@@ -10,6 +10,11 @@ import (
 	"sync"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message"
+	"github.com/sirupsen/logrus"
+)
+
+const (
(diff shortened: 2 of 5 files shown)


pkg/core/consensus/agreement/step.go
```diff
@@ -105,7 +105,10 @@ func (s *Loop) shouldCollectNow(a message.Message, round uint64, queue *consensu
 		return false
 	}
 
-	if hdr.Round > round {
+	// Only store events up to 10 rounds into the future
+	// XXX: According to protocol specs, we should abandon consensus
+	// if we notice valid messages from far into the future.
+	if hdr.Round > round && hdr.Round-round < 10 {
```
goshawk-3, line 111: Out of curiosity, have we noticed a devnet node collecting an Agreement message from  so far round? 
Much earlier, we should have this node in `OutSync` mode and Consensus/Agreement goroutines terminated.

pkg/core/consensus/queue.go
```diff
@@ -68,13 +86,15 @@ func (eq *Queue) Clear(round uint64) {
...
 	newEntries := make(map[uint64]map[uint8][]message.Message)
+	eq.items = 0
 
 	for r := range eq.entries {
 		if r > round {
 			newEntries[r] = make(map[uint8][]message.Message)
 
 			for s, m := range eq.entries[r] {
 				newEntries[r][s] = m
+				eq.items++
```
goshawk-3, line 97: Maybe this should be `eq.items += len(m)` ?

goshawk-3 (changes requested):

pkg/core/consensus/queue.go
julesdesmit: Good catch!

pkg/core/consensus/agreement/step.go
julesdesmit: Not yet, but this is possible in syncing/real network scenarios. Of course, there needs to be some proper guarding in place against maliciousness but it's part of the protocol to terminate and sync when noticing consensus messages from far in the future.

goshawk-3 (approved):
LGTM

PR #1123, opened 2021-07-14
Change quorum from 3/4ths to 2/3rds

### LINKED ISSUE #1122: Update quorum to be 2/3rds instead of 3/4ths
This implementation clashes with the protocol specs.

julesdesmit:
Fixes #1122


autholykos (approved):
LGTM

PR #1125, opened 2021-07-14
Prevent putting a transaction in the mempool twice

### LINKED ISSUE #1124: Prevent putting transactions in the mempool twice
Due to a disconnect between checking the `Contains` method, and using the `Put` method in the mempool, there is a rare edge case where a transaction can be entered into the mempool twice. It will not matter for the map, but in the `sorted` slice, having a double entry can cause the block generator to get a `nil` key, and causes a panic down the line.

julesdesmit:
Fixes #1124


autholykos (approved):
Easy enough :)

PR #1147, opened 2021-08-16
Update block finalization formula to agreement_message_{round,hash,iteration

### LINKED ISSUE #1146: Investigate culprit for potential consensus forking

goshawk-3:
fixes #1146

### DIFF
--- a/logging.md
+++ b/logging.md
@@ -35,7 +35,6 @@ grep "peer_connection" /var/log/dusk.log
 
 ## Consensus subsystem
 ```bash
-
 #  Fetch all traces from Consensus
 grep ':"consensus"'  /var/log/dusk.log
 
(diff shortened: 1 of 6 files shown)


autholykos (approved):
I think it is time to record a new issue about dropping the consensus and resync in case we begin to receive valid Agreement messages for future rounds. However, for what concerns this particular PR. it LGTM

autholykos (approved):

goshawk-3:
> I think it is time to record a new issue about dropping the consensus and resync in case we begin to receive valid Agreement messages for future rounds. However, for what concerns this particular PR. it LGTM

Having in mind that the list of active provisioners could change for future blocks, I'm not sure how a node would be capable of ensuring a future Agreement is really valid. (this is if Msg.Agreement.round > current_round + 1 )

The solution I'd envision would be to drop consensus only if we receive a fully valid next block.

PR #1153, opened 2021-08-23
Recover the Consensus loop if a successive Block acceptance fails

### LINKED ISSUE #1152: Recover the Consensus loop if  a successive Block acceptance fails
If `AcceptSuccessiveBlock` call fails due to a malicious topics.Block wire msg, this could allow an attacker to terminate the consensus loop of the majority of nodes.

```golang
// TryNextConsecutiveBlockInSync is the processing path for accepting a block
// from the network during in-sync state.
func (c *Chain) TryNextConsecutiveBlockInSync(blk block.Block, kadcastHeight byte) error {
	c.StopBlockProduction()

	if err := c.AcceptSuccessiveBlock(blk, kadcastHeight); err != nil {
		return err
	}

	return c.ProduceBlock()
}
```

goshawk-3:
fixes #1152

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -292,8 +292,8 @@ func Setup() *Server {
 		}
 	}()
 
-	if err := c.ProduceBlock(); err != nil {
-		log.WithError(err).Warn("ProduceBlock returned err")
+	if err := c.StartConsensus(); err != nil {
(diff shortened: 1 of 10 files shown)


autholykos (approved):
LGTM. I would consider extrapolating log-related information in a dedicated struct within the Chain, but it’s definitely a minor point and most properly tackled in a different issue

PR #1157, opened 2021-09-06
Mitigate Cgo overhead for ReconstructApk

### LINKED ISSUE #1162: Mitigate Cgo overhead for ReconstructApk

goshawk-3:
fixes #1162


l0k18:
Nothing important to note, just that in Go, a nil (unallocated) slice and a zero length slice are equivalent in almost every way in usage, so the `pks := make([][]byte, 0)` can be replaced with `var pks [][]byte`and you can also instead of evaluating the length of the slice, just ask if it's not nil to dispatch a set of queries.

autholykos (approved):
LGTM

PR #1159, opened 2021-09-06
Create sortition hash from Seed value

### LINKED ISSUE #1158: Include Tip.Header.Seed in deterministicSortition execution

goshawk-3:
fixes #1158


l0k18:
There is a lot of repetition of explicitly zero slice literals. Go always zeroes everything that is created, both on stack and in the heap, so `[]byte{0,0,0,0}` is equivalent to `make([]byte, 4)` and rather than repeat either of those, you could make a small function that just returns a new empty 4 member byte slice like `emptySlice(length int)` with `return make([]byte, length)`

autholykos (approved):
LGTM

PR #1161, opened 2021-09-07
Allow timer to expire before proceeding to the Reduction phase

### LINKED ISSUE #1160: Allow timer to expire before proceeding to the Reduction phase

goshawk-3:
fixes #1160


autholykos (approved):
Seems like a waste to keep waiting for the timeout when we are in the `Selection` phase and we obtained or produced the block. I understand that for now it is something that stabilizes  the node, so it is fine for now. However, I will open an issue to make sure to absorb the timeout in the reduction steps so to avert a message buildup in the `consensus.Queue`

PR #1165, opened 2021-09-14
Add RPC API to eventbus

l0k18:
In order to create a channel between independent processes, the same principle applies regardless of the nature of the transport, for this case the transport will be a net.Conn made by joining os.Stdin and os.Stdout

First task in this issue is to enumerate the set of calls and subsystems they are connected to, that they can be invoked in the same way as in-process but via a pipe. I will most likely use net/rpc library, and use a two tier structure with type name and methods grouped.

Second task will be then writing the client wrapper, which provides a rpc-bindable type with all the systems and api calls available via a convenient notation.

A note for later to check/fix, to make sure that programs that run this API that no other writing to these channels is done, rather than a concurrent-messy shared pipe, are not reading or writing to stdin/out and have independent channels for each worker process connected to this ipc-eventbus


l0k18:
Further research reveals that the components are all expecting encoded binary data and the best solution is using grpc streaming for this case, which makes this task a lot simpler. Any fail/retry logic has to be done at the client, though this should rarely happen outside of testing.

l0k18:
The end at which this work was aimed is no longer relevant, closing to clear the clutter.

PR #1177, opened 2021-10-20
Adding benchmark of agreement verification to CI

### LINKED ISSUE #1176: Add agreement verification benchmark to CI PR run
Add agreement verification benchmark to CI run

For the time being, the performance of this protocol needs to be improved so this will provide test data on each change precisely and available to all to see.

This should really be a manual job in the long run but for now it is simplest to just tack it on to the existing CI run

~~**Additional context**~~
~~Add any other context or screenshots about the feature request here.~~

l0k18:
closes #1176


l0k18:
This might be more relevant to convert to something that goes in the makefile.

PR #1183, opened 2021-11-03
Change the module path of bls12_381-sign-go

### LINKED ISSUE #1182: Redirect bls12_381-sign-go to new location inside bls12_381-sign
The places where it was previously importing bls12_381-sign-go need to be changed to bls12_381-sign.

Because the two repositories are being merged for simplicity and deduplication. https://github.com/dusk-network/bls12_381-sign/pull/34 contains the changes that will require this change.

~~**Describe alternatives you've considered**~~
~~A clear and concise description of any alternative solutions or features you've considered.~~

~~**Additional context**~~
~~Add any other context or screenshots about the feature request here.~~

l0k18:
closes #1182


pkg/core/consensus/agreement/handler.go
```diff
@@ -189,7 +191,7 @@ func verifyWhole(a message.Agreement) error {
 
 // AggregatePks reconstructs an aggregated BLS public key from a subcommittee.
 func AggregatePks(p *user.Provisioners, subcommittee sortedset.Set) ([]byte, error) {
-	if cfg.Get().Consensus.UseCompressedKeys {
+	if UseCompressedKeys {
```
goshawk-3, line 194: Any reason you rollback this line?

go.mod
```diff
@@ -83,16 +82,17 @@ require (
...
+	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/ini.v1 v1.51.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20200605160147-a5ece683394c // indirect
 )
 
 go 1.17
+
+replace google.golang.org/protobuf => google.golang.org/protobuf v1.26.1-0.20210525005349-febffdd88e85
```
goshawk-3, line 98: Should we still use replace here?

go.mod
```diff
@@ -5,7 +5,7 @@ require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/didip/tollbooth v4.0.2+incompatible
 	github.com/drewolson/testflight v1.0.0
-	github.com/dusk-network/bls12_381-sign-go v0.1.3
+	github.com/dusk-network/bls12_381-sign v0.1.0-merge2
```
goshawk-3, line 8: Shouldn't we have here `master` instead of a temp tag?

go.mod
l0k18: I think yes for the time being to reinforce the pinning of the version of protobuf. When the panic trigger is fixed later on then it would not be necessary. go.mod files have a way of mutating by themselves and a replace won't be mutated.

pkg/core/consensus/agreement/handler.go
l0k18: This would have been unintentional, if you don't mind to give a double check there isn't other erroneous changes that would be great.

go.mod
l0k18: I don't think the master branch is the same as fix-33 because we were waiting to confirm all of this. So when master branch is also the merged together ipc/cgo combo version it would be changed to `latest` and it would grab from the master branch.

l0k18: https://github.com/dusk-network/bls12_381-sign/pull/34 I have noted that we should merge that PR before this one so as to make the new merged repository the master/release state of it. Then this repo's go.mod should be amended to the latest on the other repo now master branch and this one can then be merged and closed.

herr-seppia:
Closing in favour of #1203

PR #1199, opened 2021-11-17
Implement fallback procedure upon blockchain fork corner case

### LINKED ISSUE #1198: Implement fallback procedure upon blockchain fork corner case
Upon arrival of a block with height equal to local blockchain tip, we need to ensure if this is due to recent consensus-split occurrence. If that's confirmed, a fallback procedure as a recovery  should  be triggered.

The fallback should be capable to:
- Rewrite Blockchain tip (both on-disk and in-memory)
- Restore Contract Storage state
- Resubmit missing Mempool transactions
- Restore Wallet state (if affected)

Even when the consensus split issue gets solved, we should still play safe and implement a fallback procedure so to eliminate the chance of a stalled network. 

**Additional context**
This is deemed as a candidate solution for #1156

goshawk-3:
fix #1198

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -63,12 +63,12 @@ type Server struct {
 
 // LaunchChain instantiates a chain.Loader, does the wire up to create a Chain
 // component and performs a DB sanity check.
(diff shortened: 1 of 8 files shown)


autholykos (approved):
LGTM

PR #1205, opened 2021-11-26
Integrate rusk state grpc service to both node and rusk mock

### LINKED ISSUE #1201: Integrate rusk state grpc service to both node and rusk mock

- Incorporate **state.Accept**, **state.Finalized**, **state.GetEphemeralStateRoot**, **state.VerifyStateTransition**, **state.ExecuteStateTransition**  rusk.schema APIs.
-  Update Mock handlers for new APis in Rusk mock.
-  Update Node.Proxy
- Implement **state_root** hash in Block Header as a hashable field
- Initialize **state_root** at generating block candidate phase
- Introduce **Finalized** and **Tentative block**
- Ensure both node and rusk services are on the same **state_root**
 

- BlockGenerator  must ensure candidate block transactions are valid according to Rusk most recent state.
- BlockGenerator  must ensure that cumulative gas of candidate block transactions does not exceed **MaxBlockGasLimit**
- Both (node and rusk) services must be synced up based on **state_root** value.
- Node must **finalize** a block at first consensus iteration only.

### LINKED ISSUE #1202: Add TxRoot to Hashable Block Header fields
**Describe the bug**
TxRoot Block Header value is not part of Hashable fields.  This allows an attacker to append transactions to an accepted block bypassing block header validation.

**To Reproduce**
See also [code](https://github.com/dusk-network/dusk-blockchain/blob/a6d9b65e988e6e1d7079b99863f10d52e928eea6/pkg/core/data/block/header.go#L113)
 

**Expected behaviour**
Block hash should be based on TxRoot.
 
**Additional context**

goshawk-3:
fixes #1201 
fixes #1202


autholykos (approved):
I trust the specs have been followed (in reference to the hash being 32 bytes and the Block Gas Limit being specified by the Node). Some more commenting would be nice, but I think the code is solid.

PR #1206, opened 2021-12-01
Implement priority queue on inbound messages

### LINKED ISSUE #1187: Implement priority queue on inbound messages
A queue that sidesteps the non-determinism of golang scheduling system: https://github.com/dusk-network/dusk-blockchain/blob/f44e1055887cf2a71fd05ec346fc3461cf223d44/pkg/p2p/peer/peer.go#L340

There is a strong suspicion that some consensus messages might be unlawfully delayed by the golang scheduler, potentially creating problems.

Splitting the inbound messages between a priority queue and the normal scheduler as is

**Additional context**
This might be related to the root cause of #1156

l0k18:
A priority queue which is designed to attempt to focus primarily on priority items without neglecting non-priority items completely. The implementation achieves this by two rules:

- If there is more items in either priority or non-priority, first consume the ones that are out of balance
- Every time a whole factor of `factor` is in the count of one of the queues, it selects that queue automatically.

I'm not sure what value `factor` should be set at, something like 3 or 4. It's primary function is to make a rule that unconditionally will empty the non-overbalanced queue at least to some degree and not let it get too stale.

fixes #1187


l0k18:
The current state of this, in brief, is that it now properly forces selection of an available non-priority message at a cycle repeating by the `factor`, if it would otherwise select a priority item, as a means to stop nonpriority messages getting flooded out.

The final step is to put the queue into action. The existing code simply spawns arbitrary goroutines to deal with processes, instead we will run one goroutine that pops new messages off.

note for future: If parallelism is required it can be added later and also categorize by message types to determine where to dispatch the message.

l0k18:
As the implementation is being put to use, currently the reader message queue processing goroutine has to poll for messages if it uses the `Pop()` function as it is currently written. A better way needs to be made for the no message case so the pop function blocks on it, as it should, to an empty queue.

l0k18:
The dangling element to close this issue is that in the test the callback fails to execute, for which reason a timeout failure mode was added.

goshawk-3:
Due to inactivity on this, the PR is out of date and should be closed. However the branch should be kept for future reference.

PR #1210, opened 2021-12-14
Integrate Kadcast gRPC client in the node

### LINKED ISSUE #1207: Implement client to kadcast gRPC interface of rusk network service
Instead of handle kadcast peers internally, the node shall use the gRPC service exposed by rusk as per dusk-network/rusk#380

Similarily to p2p/kadcast package, we should implement a new p2p package (p2p/kadcast_client) that implements Writer and Reader structs:

**Writer** struct should:

- Handle **topics.Kadcast** messages from **EventBus**
- Builds a grpc kadcast message
- Send grpc.Broadcast 

**Reader** struct should:

- Starts listening for new messages (grpc.Listen)
- On new message, it should delegate it to **peer.MessageProcessor.Collect**)

Optional here is to integrate the **new priority queue**.

We want to use the new rust kadcast library, and the gRPC communication seems the best choice as per dusk-network/rusk#380

Keep going with the internal kadcast impl

**Additional context**
The list of topics which will use the kadcast network is still under review.

### LINKED ISSUE #1222: Deprecate Golang Kadcast package
**Describe the bug**
With regard to positive results from stress testing of Rust Kadcast, we should move on and remove or update accordingly:

- `dusk-network/dusk-blockchain/pkg/p2p/kadcast`
- kadcast launch from dusk.server
- any kadcast-related configuration
- Move `/pkg/p2p/kadcli` to `/pkg/p2p/kadcast`

t00ts:
Implements a gRPC client for the network layer (kadcast) provided by Rusk.

Fixes #1207 
Fixes #1222


pkg/p2p/kadcli/reader.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/dusk-network/dusk-protobuf/autogen/go/rusk"
+)
+
+// Reader abstracts all of the logic and fields needed to receive messages from
+// other network nodes.
+type Reader struct {
+	processor *peer.MessageProcessor
+	gossip    *protocol.Gossip
+
+	cli  rusk.NetworkClient
```
goshawk-3, line 30: Please rename this to `client` as cli might be confused with CLI (Command Line Interface). Once we deprecate the `kadcast` golang impl, we'll rename the package from kadcli to `kadcast`

pkg/p2p/kadcli/reader.go
```diff
@@ -0,0 +1,120 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				// receive a message
+				msg, err := stream.Recv()
+				if err == io.EOF {
+					// TODO: Notify someone? Retry mechanism?
+					return
```
goshawk-3, line 67: Let's report an warning message here so we can detect this on testing.

pkg/p2p/kadcli/reader.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// processMessage propagates the received kadcast message into the event bus.
+func (r *Reader) processMessage(message *rusk.Message) {
+	reader := bytes.NewReader(message.Message)
+
+	// read message (extract length and magic)
+	b, err := r.gossip.ReadMessage(reader)
+	if err != nil {
+		log.WithError(err).Warnln("error reading message")
```
goshawk-3, line 85: Anytime we report an error\warning  in `processMessage` we should add `WithField("r_addr", message.Metadata.SrcAddress)` so that we can trace back the malformed message.

pkg/p2p/kadcli/writer.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil
+}
+
+// sendPacket passes a message to the kadkast peer to be sent to a peer.
+func (w *Writer) sendPacket(addr string, payload []byte) error {
+	m := &rusk.SendMessage{
+		TargetAddress: addr,
+		Message:       payload,
+	}
+	if _, err := w.cli.Send(context.TODO(), m); err != nil {
```
goshawk-3, line 123: We should use the same `stop` context in both sendPacket and broadcast.

cmd/dusk/server.go
```diff
@@ -374,10 +390,11 @@ func setupGRPCClients(ctx context.Context) (transactions.Proxy, *grpc.ClientConn
 	keysClient, _ := client.CreateKeysClient(ctx, addr)
 	transferClient, _ := client.CreateTransferClient(ctx, addr)
 	stakeClient, _ := client.CreateStakeClient(ctx, addr)
+	networkClient, _ := client.CreateNetworkClient(ctx, addr)
```
goshawk-3, line 393: This is not needed as there is another Network client connection  in `Launch`

```
func (p *Peer) Launch(conn *grpc.ClientConn) {
	// gRPC rusk client
	ruskc := rusk.NewNetworkClient(conn)
```

pkg/p2p/kadcli/peer_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// create a mock message
+	buf, err := createBlockMessage()
+	if err != nil {
+		t.Errorf("fail to create msg: %v", err)
+	}
+
+	go func() {
+		// send a broadcast message
+		if errw := w.WriteToAll(buf.Bytes(), []byte{127}, 0); errw != nil {
```
goshawk-3, line 115: We should use `eventbus.Publish(topics.Kadcast, ...`  so that we can ensure both subscription and message handling are working properly.

pkg/p2p/kadcli/writer.go
```diff
@@ -0,0 +1,135 @@
...
+			log.WithError(err).Warn("write failed")
+		}
+	}()
+	return 0, nil
+}
+
+// WriteToAll broadcasts message to the entire network.
+// The kadcast height is read from message Header.
+// Note: Assumes the message is properly encoded (no pre-processing done here).
+func (w *Writer) WriteToAll(data, header []byte, priority byte) error {
```
goshawk-3, line 78: This should be private.

pkg/p2p/kadcli/writer.go
```diff
@@ -0,0 +1,135 @@
...
+	// exctract kadcast height
+	height := header[0]
+	// broadcast
+	return w.broadcastPacket(height, data)
+}
+
+// WriteToPoint writes a message to a single destination.
+// The receiver address is read from message Header.
+// Note: Assumes the message is properly encoded (no pre-processing done here).
+func (w *Writer) WriteToPoint(data, header []byte, priority byte) error {
```
goshawk-3, line 92: This should be private.

pkg/p2p/kadcli/reader.go
```diff
@@ -0,0 +1,120 @@
...
+		cli:       rusk,
+	}
+}
+
+// Listen starts accepting and processing stream data.
+func (r *Reader) Listen() {
+	// create stream handler
+	stream, err := r.cli.Listen(context.Background(), &rusk.Null{})
+	if err != nil {
+		log.Fatalf("open stream error %v", err)
```
goshawk-3, line 48: Let's use `WithError` here

pkg/p2p/kadcli/reader.go
```diff
@@ -0,0 +1,121 @@
...
+	// verify checksum
+	if !checksum.Verify(msg, cs) {
+		log.WithError(errors.New("invalid checksum")).
+			Warnln("error verifying message cs")
+		return
+	}
+
+	// collect (process) the message
+	go func() {
+		if _, err = r.processor.Collect(message.Metadata.SrcAddress, msg, nil, protocol.FullNode, nil); err != nil {
```
goshawk-3, line 104: Keep in mind that we need to turn the responses into `KadcastPoint` messages otherwise syncing will not work.

```golang

// TODO: set service flag properly
	respBufs, err := r.processor.Collect(raddr, m, nil, protocol.FullNode, []byte{height})
	if err != nil {
		ll.WithError(err).Error("messageProcessor failed to collect message")
	}

	// Any response message is translated into Kadcast Point-to-Point wire message.
	// In other words, any bufs item is sent back to the sender node (remotePeer).
	for i := 0; i < len(respBufs); i++ {
		ll.WithField("raddr", remotePeer.Address()).Trace("send point-to-point message")

		// Send Kadcast point-to-point message with destination raddr
		r.publisher.Publish(topics.KadcastPoint,
			message.NewWithHeader(topics.KadcastPoint, respBufs[i], []byte(remotePeer.Address())))
	}
```

goshawk-3, line 104: We should not forget to pass `height` value to upper layers.

goshawk-3 (approved):
LGTM

PR #1221, opened 2021-12-23
Create a testbed package to stress test kadcast protocol

### LINKED ISSUE #1219: Implement a testbed to stress test new kadcast rust
A test unit (TestCluster) that bootstraps a Kadcast network on localhost with configurable size.  It creates a grpc client per each node and also performs a message broadcast to the network.  It asserts that a broadcasted message is always received by all network active nodes.

Additionally, it could measure the TPS capabilities of Kadcast network.

Testbed will be used to stress test the Rust Kadcast impl. A few factors to be taken into account - cluster_size, message_size and network_latency. Later on, it could be also a handful platform to check for regressions on releasing a new Kadcast version.

Using test-harness as a testbed for Kadcast would be inefficient and can make issues debugging cumbersome.

goshawk-3:
fixes #1219

### DIFF
--- a/pkg/p2p/testbed/cluster_test.go
+++ b/pkg/p2p/testbed/cluster_test.go
@@ -1,0 +1,153 @@
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT License was not distributed with this
+// file, you can obtain one at https://opensource.org/licenses/MIT.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+//go:build testbed
+// +build testbed
+
+package testbed
+
+import (
+	"context"
+	"encoding/hex"
+	"io"
+	"os/exec"
+	"strconv"
+	"sync"
+
+	"github.com/dusk-network/dusk-protobuf/autogen/go/rusk"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+)
+
+type testNode struct {
+	kadcastPort int
+	grpcPort    int
+
+	exe *exec.Cmd
+
+	mu      sync.RWMutex
+	message []byte
+}
+
+func newNode(exePath string, bootstrapNodes []string, kadcastPort, grpcPort int) (*testNode, error) {
+	n := &testNode{
+		kadcastPort: kadcastPort,
+		grpcPort:    grpcPort,
+		message:     make([]byte, 0),
+	}
+
+	cmd := exec.Command(exePath,
+		"--ipc_method",
+		"tcp_ip",
+		"-p",
+		strconv.Itoa(grpcPort),
+		"--kadcast_public_address",
+		NetworkAddr+":"+strconv.Itoa(kadcastPort),
+		"--kadcast_bootstrap",
+		bootstrapNodes[0],
+		"--kadcast_bootstrap",
+		bootstrapNodes[1],
+	)
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	log.WithField("kadcastPort", kadcastPort).
+		WithField("grpc", grpcPort).
+		Info("Start a node")
+
+	n.exe = cmd
+
+	return n, nil
+}
+
+// GetMessage returns a copy of a received msg.
+func (t *testNode) GetMessage() []byte {
+	t.mu.RLock()
+	cpy := make([]byte, len(t.message))
+	copy(cpy, t.message)
+	t.mu.RUnlock()
+
+	return cpy
+}
+
+// Kill causes the Rusk Process to exit immediately.
+func (t *testNode) Kill() {
+	if t.exe != nil {
+		t.exe.Process.Kill()
+	}
+}
+
+// Broadcast sends broadcast grpc call from this node.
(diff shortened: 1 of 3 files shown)


pkg/p2p/testbed/kadcast_test.go
```diff
@@ -0,0 +1,114 @@
...
+	for i := 0; i < size; i++ {
+		n, err := newNode(ruskExecutable, bootstrapNodes, baseKadcastPort+i, baseGRPCPort+i)
+		if err != nil {
+			t.Error(err)
+		}
+
+		switch i {
+		case 1:
+			// Allow bootstrappers to start up
+			time.Sleep(5 * time.Second)
```
l0k18, line 58: Using delays to account for startup is ok for tests but can be unnecessarily slow and occasionally too short, a signal is better. A super simple method would be to add a ready signal in the form of an empty file in a common-accessible area between the apps (can even be in tmp), and change this loop to pause a few hundred milliseconds and then check for the existence of the file, and repeat until a timeout.

l0k18: A possible signal mechanism could be the lock file of databases used by Rusk, if this can be easily located and checked by dusk-blockchain

goshawk-3: No signal mechanism is available for Kadcast bootstrapping and this is simply a test.

pkg/p2p/testbed/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"context"
+	"encoding/hex"
+	"io"
+	"os/exec"
+	"strconv"
+	"sync"
+
+	"github.com/dusk-network/dusk-protobuf/autogen/go/rusk"
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
```
t00ts, line 22: Any reason for the `logrus` import to be both named and unnamed?

pkg/p2p/testbed/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	conn, _ := createNetworkClient(ctx, addr)
+
+	// create stream handler
+	stream, err := conn.Listen(context.Background(), &rusk.Null{})
+	if err != nil {
+		log.WithError(err).Error("open stream error")
+		return
+	}
+
+	logrus.WithField("kadcast_port", t.kadcastPort).Info("Listening")
```
t00ts, line 128: Again, curious about the reason for some logs to use `logrus.[...]` and others use `log.[...]`

pkg/p2p/testbed/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// accept  stream messages
+	for {
+		msg, err := stream.Recv()
+		if err != nil {
+			if err == io.EOF {
+				return
```
t00ts, line 135: Should we have a warning here, or not yet?

pkg/p2p/testbed/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	return rusk.NewNetworkClient(conn), conn
+}
+
+// Listen creates a client connection and accepts stream messages.
+func (t *testNode) Listen(ctx context.Context) {
+	addr := NetworkAddr + ":" + strconv.Itoa(t.grpcPort)
+	conn, _ := createNetworkClient(ctx, addr)
+
+	// create stream handler
+	stream, err := conn.Listen(context.Background(), &rusk.Null{})
```
t00ts, line 122: Should this use the the `ctx` that's passed as general context as well?

pkg/p2p/testbed/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	for {
+		msg, err := stream.Recv()
+		if err != nil {
+			if err == io.EOF {
+				return
+			}
+
+			logrus.WithError(err).Error("recv error")
+			return
+		}
```
t00ts, line 140: Should we add a `select` block within the `for` loop to check for the `ctx.Done()` signal as well?

t00ts (approved):

pkg/p2p/testbed/cluster_test.go
goshawk-3: Good catch!

pkg/p2p/testbed/cluster_test.go
goshawk-3: Info should be enough for the context of testbed

pkg/p2p/testbed/cluster_test.go
goshawk-3: definitely a typo.

pkg/p2p/testbed/cluster_test.go
goshawk-3: typo

pkg/p2p/testbed/cluster_test.go
goshawk-3: Not needed. If you cancel the ctx that is passed to Listen, it will close the stream with `error="rpc error: code = Canceled desc = context canceled"`

PR #1233, opened 2022-01-07
Support dial timeout and unix domain network for kadcast grpc client

### LINKED ISSUE #1232: Support  dial timeout and unix domain network for kadcast grpc client
**Describe the bug**
For efficiency both rusk and dusk should use unix domain as IPC by default. In addition, Dial Timeout should be supported so that dusk could wait for rusk service to finalize its startup procedure.

Following configurations to be added:

```toml
# Kadcast peer settings
[kadcast]
# if disabled, gossip protocol is active
enabled=false

# grpc client connection config
[kadcast.grpc]
network="tcp"
address="127.0.0.1:8585"
# Number of seconds to wait for client conn establishment
dialTimeout = 10

```

goshawk-3:
fixes #1232

### DIFF
--- a/harness/engine/network.go
+++ b/harness/engine/network.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


herr-seppia (approved):
LGTM

PR #1235, opened 2022-01-11
Disallow ChanListeners to report warnings on a suspended consensus

### LINKED ISSUE #1149: Solve message channel buffer is full warning
While synchronizing, the inbound peer message buffer complains about being full.

When a node discoveres its copy of the blockchain is out of date, the consensus processing stops, but peers don't stop sending messages that would be handled by the consensus loop. 

Examine the message consumers and give them an offline mode so the messages are discarded in some kind of timely fashion.

goshawk-3:
fixes #1149

### DIFF
--- a/pkg/util/nativeutils/eventbus/listener.go
+++ b/pkg/util/nativeutils/eventbus/listener.go
@@ -11,7 +11,9 @@ import (
 	"errors"
 	"math/big"
 	"sync"
+	"sync/atomic"
 
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/p2p/wire/message"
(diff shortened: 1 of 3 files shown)


pkg/util/nativeutils/eventbus/listener.go
```diff
@@ -20,10 +22,22 @@ import (
...
+	ErrRingBufferClosed = errors.New("ringbuffer is closed")
+)
+
 // Listener publishes a byte array that subscribers of the EventBus can use.
 type Listener interface {
 	// Notify a listener of a new message.
 	Notify(message.Message) error
+
+	// Update Listeners log level. From verbose to silent.
+	SetLogLevel(logrus.Level)
```
autholykos, line 39: Replace the `SetLogLevel` method with `Setup` and `Teardown`. The `LogLevel` can be manipulated in these methods rather than being exposed. See the other comments above

autholykos (changes requested):
See my comments. I would prefer the `setup` and `teardown` in the consensus to propagate to a `Listener.Setup` and `Listener.Teardown` so to centralize log (and potentially other) configuration management there

pkg/util/nativeutils/eventbus/listener.go
goshawk-3: It makes sense to me to replace `SetLogLevel` with `Setup(loglevel)` method but not sure there is any need for `Listener.Teardown`. At consensus.Teardown method we need to update  ChanListener to make it temporarily silent but not to tear it down.

PR #1236, opened 2022-01-11
Enable test-harness running rusk with unix domain socket IPC

### LINKED ISSUE #1234: Use UDS to communicate with rusk
Use UDS instead of TCP to communicate with rusk gRPC services

UDS is faster then TCP protocol

**Additional context**
Actually there is a bug in `h2` rusk dependency that results in `Stream terminated by RST_STREAM with PROTOCOL_ERROR` if no authority is specified

To avoid this you can apply this kind of workaround

* For grpcurl clients: add `-authority "dummy"`
* For grpc-go clients: use `grpc.Dial("unix:///blah.sock", grpc.WithAuthority("dummy"), grpc.WithInsecure())`

goshawk-3:
fixes #1234

### DIFF
--- a/harness/engine/network.go
+++ b/harness/engine/network.go
@@ -15,7 +15,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


herr-seppia (approved):
LGTM

Issue #1059, opened 2021-05-31
Implement a `diskpool` for transaction storage


julesdesmit:
This will avoid potential OOM-reaper activations when running nodes with limited RAM.

autholykos:
@julesdesmit is this still relevant? I remember that we implemented something like that when battling against the OOM issue

autholykos:
To be assessed in the context of Testnet

goshawk-3:
pendingPropagation queue should be also persisted on mempool closing.

PR #1238, opened 2022-01-13
Support both rate limiter and pendingPropagation in mempool

### LINKED ISSUE #1237: Implement back pressure on mempool tx processing
We should limit the rate at which mempool verifies and respectively re-propagates messages. 

Example results:

Recently noticed on test-harness with 20 provisioners,  block time increased 10 times when all 20 nodes start sending txs.

```
level=info msg="Measured TPS 0.10 (1/10)" average_block_time=4 cumulative_txs_count=33 height=22
level=info msg="Measured TPS 0.08 (1/12)" average_block_time=5 cumulative_txs_count=34 height=23
level=info msg="Measured TPS 0.07 (1/15)" average_block_time=5 cumulative_txs_count=35 height=24

...

level=info msg="Measured TPS 24.67 (2886/117)" average_block_time=13 cumulative_txs_count=2923 height=27
level=info msg="Measured TPS 56.59 (2886/51)" average_block_time=15 cumulative_txs_count=5809 height=28
level=info msg="Measured TPS 11.20 (224/20)" average_block_time=15 cumulative_txs_count=6033 height=29
```

```
{"hash":"e9cc013fe6...4e224c6b43","level":"info","msg":"consensus_achieved","phase":"agreement","process":"consensus","round":27,"short-circuit":true,"step":12,"stepvotes":[{"BitSet":4060,"Signature":"hIwVwq9cheyVw2BpSzHCMcta3Vb5UAXObBpnK8kLQlf32V2MGu5RVcXX+Age8+YF"},{"BitSet":6051,"Signature":"r6xnV935oDLqIFjmpKSCkfqTe8svsDlMsRW9oyZjrCp0unL9+dx6DO68xGZA6lPm"}]}
{"hash":"2d94730a91...e80c952e76","level":"info","msg":"consensus_achieved","phase":"agreement","process":"consensus","round":28,"short-circuit":true,"step":6,"stepvotes":[{"BitSet":15586,"Signature":"lfqGBZt5YlBluE9mXMDnTR+mzok0uKo74YbQ26iFH2VHN2WuyESgd+Rw0j14sKB0"},{"BitSet":12703,"Signature":"thX22lgxYAN6TaVKCtGF+Iu8bVhkRendy6Mf1RZ+tBz638lK0+I1FIS319oFcS6O"}]}


```

This would allow Network consensus to maintain a steady block time and mainly finalize a block at first iteration. This should also mitigate the pressure on Rusk grpc.VerifyTransaction.

goshawk-3:
fixes #1237

### DIFF
--- a/pkg/core/mempool/mempool.go
+++ b/pkg/core/mempool/mempool.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/util/diagnostics"
+	"golang.org/x/time/rate"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/config"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
@@ -55,6 +56,8 @@ type Mempool struct {
 	// verified txs to be included in next block.
 	verified Pool
 
+	pendingPropagation chan TxDesc
+
 	// the collector to listen for new accepted blocks.
 	acceptedBlockChan <-chan block.Block
 
@@ -65,6 +68,8 @@ type Mempool struct {
 
 	// the magic function that knows best what is valid chain Tx.
 	verifier transactions.UnconfirmedTxProber
+
+	limiter *rate.Limiter
 }
 
 // checkTx is responsible to determine if a tx is valid or not.
@@ -103,6 +108,19 @@ func NewMempool(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier tra
 
 	acceptedBlockChan, _ := consensus.InitAcceptedBlockUpdate(eventBus)
 
+	// Enable rate limiter from config
+	fromConfig := config.Get().Mempool.PropagateTimeout
+
+	var limiter *rate.Limiter
+	if len(fromConfig) > 0 {
+		timeout, err := time.ParseDuration(config.Get().Mempool.PropagateTimeout)
+		if err != nil {
+			log.WithError(err).Fatal("could not parse mempool propagation timeout")
+		}
+
+		limiter = rate.NewLimiter(rate.Every(timeout), 1)
+	}
+
 	m := &Mempool{
 		eventBus:                eventBus,
 		latestBlockTimestamp:    math.MinInt32,
@@ -111,6 +129,8 @@ func NewMempool(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier tra
 		getMempoolTxsBySizeChan: getMempoolTxsBySizeChan,
 		sendTxChan:              sendTxChan,
 		verifier:                verifier,
+		limiter:                 limiter,
+		pendingPropagation:      make(chan TxDesc, 1000),
 	}
 
 	// Setting the pool where to cache verified transactions.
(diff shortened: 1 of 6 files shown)


pkg/core/mempool/mempool.go
```diff
@@ -103,6 +108,20 @@ func NewMempool(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier tra
...
+
+	var limiter *rate.Limiter
+
+	if len(fromConfig) > 0 {
+		timeout, err := time.ParseDuration(config.Get().Mempool.PropagateTimeout)
+		if err != nil {
+			log.WithError(err).Fatal("could not parse mempool propagation timeout")
+		}
+
+		limiter = rate.NewLimiter(rate.Every(timeout), 1)
```
autholykos, line 122: Burst of 1 every 100ms sounds quite low. Maybe consider a higher value (2 or 3 without message priority, higher with priority)

pkg/core/mempool/mempool.go
```diff
@@ -126,42 +154,80 @@ func NewMempool(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier tra
...
+				err = m.advertiseTx(txid)
+			}
+
+			if err != nil {
+				log.WithField("txid", hex.EncodeToString(txid)).WithError(err).Error("failed to propagate")
+			}
+
+		// Mempool terminating
+		case <-ctx.Done():
+			return
```
autholykos, line 225: Shouldn't we drain `m.pendingPropagation` in case of context cancelation?

pkg/core/mempool/mempool.go
```diff
@@ -126,42 +154,80 @@ func NewMempool(eventBus *eventbus.EventBus, rpcBus *rpcbus.RPCBus, verifier tra
...
+func (m *Mempool) propagateLoop(ctx context.Context) {
+	for {
+		select {
+		case t := <-m.pendingPropagation:
+			// Ensure we propagate at proper rate
+			if m.limiter != nil {
+				if err := m.limiter.Wait(ctx); err != nil {
+					log.WithError(err).Error("failed to limit rate")
+				}
+			}
```
autholykos, line 203: The ordering of Transactions to be propagated should follow an order based on the `Fee` field. This needs to be tackled in a different PR

autholykos (changes requested):
Some minor changes requested: 
- If my understanding of the rate limiting is correct, we should probably allow an additional configuration parameter to tune transaction burst other than `1`.
- Context cancelation leakage should be assessed. It might be that the `m.pendingPropagation` channel would be garbage collected, but this needs to be double checked

pkg/core/mempool/mempool.go
goshawk-3: Not for now probably. Later on when we implement #1059 , we can drain and persist these txs as well.

pkg/core/mempool/mempool.go
goshawk-3: I'll add it. Could be good to play with it on benchmark tests too.

pkg/core/mempool/mempool.go
goshawk-3: FYI
https://github.com/dusk-network/dusk-blockchain/issues/1059#issuecomment-1012049540

pkg/core/mempool/mempool.go
autholykos: Referenced in #1240

autholykos (approved):
LGTM

PR #1242, opened 2022-01-13
Reintroduce priority mapper in topics.Kadcast Collector

### LINKED ISSUE #1239: Re-enable sort-by-priority for kadcast outbound messages
**Describe the bug**
We should reintroduce priorities on ring buffer that collects Kadcast outbound messages.

goshawk-3:
fixes #1239

### DIFF
--- a/pkg/p2p/kadcast/writer.go
+++ b/pkg/p2p/kadcast/writer.go
@@ -17,6 +17,11 @@ import (
 	"github.com/dusk-network/dusk-protobuf/autogen/go/rusk"
 )
 
+const (
+	// MaxWriterQueueSize max number of messages queued for broadcasting.
+	MaxWriterQueueSize = 1000
+)
+
 // Writer is a proxy between EventBus and Kadcast service. It subscribes for
 // both topics.Kadcast and topics.KadcastPoint, compiles a valid wire frame and
 // propagates the message to Kadcast service.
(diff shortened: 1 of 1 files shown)


pkg/p2p/kadcast/writer.go
```diff
@@ -55,26 +76,25 @@ func (w *Writer) Subscribe() {
...
-		}
-		// broadcast a message
-		if len(header) == 1 {
-			err = w.writeToAll(data, header, priority)
-		}
-		// log errors
-		if err != nil {
-			log.WithError(err).Warn("write failed")
-		}
-	}()
```
autholykos, line 77: I suppose removing the goroutine has to do with the fact that the `Consumer` already consumes the messages concurrently

autholykos (approved):
LGTM

herr-seppia (approved):

pkg/p2p/kadcast/writer.go
goshawk-3: Both writeToAll/writeToPoint may block due to grpc client request they call. When this happens, runtime starts queue-ing goroutines. However our purpose with this PR is to queue messages in the Listener Ring buffer so that they are always sorted by priority/(type of topic).

If that makes sense to you @autholykos ?

autholykos: Yes it does

PR #1244, opened 2022-01-13
Allow to launch testbed on a remote cluster

### LINKED ISSUE #1243: Remote testbed
Allow to launch `testbed` which connect to remote rusk instances instead of spawning local cluster

This way a more precise stress result can be performed with real latency/drop introduced

herr-seppia:
Use env CLUSTER_CSV=<your_csv_path> to use it

Resolves #1243


goshawk-3 (approved):
LGTM

PR #1245, opened 2022-01-14
Terminate gracefully all subsystems/components and pending grpc requests

### LINKED ISSUE #1241: Implement graceful shutdown for components
**Describe the bug**
On recent review, It was noticed that dusk.Server does not cancel the context passed to Mempool, Chain and GRPC clients.

**Expected behaviour**
On calling `func (s *Server) Close()`,  cancelling the Parent(Server) context is a must. Also, we could revise if Parent Context can be used for closing other components instead of explicit `Close` call.

goshawk-3:
fixes #1241 

- Cancel all goroutine loops (kadcast listener, consensus, mempool, chain, gossip read loop)
- Close non-cancelable components (db driver, gql server, eventbus)
- Cancel pending grpc client requests on state/network services.

### DIFF
--- a/cmd/dusk/server.go
+++ b/cmd/dusk/server.go
@@ -231,6 +231,7 @@ func Setup() *Server {
 
 	// Instantiate GraphQL server
 	var gqlServer *gql.Server
+
 	if cfg.Get().Gql.Enabled {
 		var e error
 		if gqlServer, e = gql.NewHTTPServer(eventBus, rpcBus); e != nil {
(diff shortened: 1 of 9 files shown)


cmd/dusk/server.go
```diff
@@ -299,16 +315,36 @@ func Setup() *Server {
...
+	// Close graphql server.
+	if s.gqlServer != nil {
+		if err := s.gqlServer.Close(); err != nil {
+			log.WithError(err).Warn("failed to close gql server")
+		}
+	}
+
+	if s.grpcServer != nil {
+		s.grpcServer.GracefulStop()
+	}
```
autholykos, line 330: Does the golang process still need to serve any data over gRPC?

cmd/dusk/server.go
```diff
@@ -299,16 +315,36 @@ func Setup() *Server {
...
 	}
+
+	if s.dbDriver != nil {
+		if err := s.dbDriver.Close(); err != nil {
+			log.WithError(err).Warn("failed to close db driver")
+		}
+	}
+
+	s.rpcBus.Close()
+	s.eventBus.Close()
```
autholykos, line 347: Some of these closures should be done in parallel using a `WaitGroup` to speed up the teardown process

autholykos (approved):
There are some minor leftovers for `Server` termination to be tackled by other issues. LGTM

herr-seppia (approved):

cmd/dusk/server.go
goshawk-3: It's going to be removed on merging testnet branch. For now we need gRPC for maintaining healthy test-harness. Test-harness pushes stake and transfer transactions via grpc.

PR #1248, opened 2022-01-14
Remote testbed verify truncated message

### LINKED ISSUE #1247: Remote testbed shouldn't download the whole message
Remote Network.Listen should expect a truncated message 

According to dusk-network/rusk#431 the message is truncated to first 1000bytes for tests

herr-seppia:
Resolves #1247

### DIFF
--- a/pkg/p2p/testbed/kadcast_test.go
+++ b/pkg/p2p/testbed/kadcast_test.go
@@ -118,18 +118,18 @@ func connectToRemoteCluster(ctx context.Context, t *testing.T) []*testNode {
 	return cluster
 }
 
-func connectToCluster(ctx context.Context, t *testing.T) []*testNode {
+func connectToCluster(ctx context.Context, t *testing.T) ([]*testNode, bool){
 
 	if ruskCsvPath == "" {
-		return bootstrapCluster(ctx, t)
+		return bootstrapCluster(ctx, t), false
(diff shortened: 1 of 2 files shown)


autholykos (approved):
LGTM

pkg/p2p/testbed/kadcast_test.go
```diff
@@ -139,6 +139,9 @@ func assertBroadcastMsgReceived(t *testing.T, cluster []*testNode, sender int, d
 	blob, _ := crypto.RandEntropy(uint32(msgSize))
 	cluster[sender].Broadcast(context.Background(), blob)
 
+	if is_remote {
+		blob = blob[:1000]
```
goshawk-3, line 143: If we need to run a smoke test with payload size of 100bytes (less than 1000) this will panic.

goshawk-3, line 143: Any particular reason why message truncation is chosen here instead of  a simple 32 hash of the payload? For integrity, I'd use hash.

goshawk-3 (changes requested):

pkg/p2p/testbed/kadcast_test.go
herr-seppia: Moved to blake2b 256hash

PR #1259, opened 2022-01-24
Support blocktimestamp field in transactions query

### LINKED ISSUE #1252: Implementing Timestamp on gql endpoint
We need to include transaction timestamp to the graphql endpoints. This will allow the field to be fetched with the single tx queries.
We can use the block timestamp for this purpose, for the time being.

Transaction timestamp is a required field for the block explorer.

We could cross fetch it with two queries (getting the tx first to get the block hash, and then querying the block hash to get the timestamp), but this would greatly degrade performance.

**Additional context**
n/a

goshawk-3:
fixes #1252

### DIFF
--- a/pkg/gql/README.md
+++ b/pkg/gql/README.md
@@ -141,7 +141,7 @@ NB: The examples from below represent only query structures. To send a query as 
   }
   ```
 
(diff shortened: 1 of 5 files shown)


autholykos (approved):
LGTM

PR #1272, opened 2022-01-26
Generate, store and load Consensus keys

### LINKED ISSUE #1208: Persist and manage Consensus keys separately from Wallet keys
The golang process needs to generate, store and load the consensus keys

goshawk-3:
fixes #1208

### DIFF
--- a/cmd/dusk/loadkeys.go
+++ b/cmd/dusk/loadkeys.go
@@ -1,0 +1,119 @@
+// This Source Code Form is subject to the terms of the MIT License.
+// If a copy of the MIT License was not distributed with this
+// file, you can obtain one at https://opensource.org/licenses/MIT.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+package main
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+
+	cfg "github.com/dusk-network/dusk-blockchain/pkg/config"
+	consensuskey "github.com/dusk-network/dusk-blockchain/pkg/core/consensus/key"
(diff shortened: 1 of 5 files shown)


autholykos (approved):
LGTM

herr-seppia (approved):
LGTM

cmd/dusk/loadkeys.go
```diff
@@ -0,0 +1,119 @@
...
+			if err := keys.Save(passwd, filePath); err != nil {
+				return consensuskey.Keys{}, err
+			}
+
+			log.WithField("pubkey", util.StringifyBytes(keys.BLSPubKey)).Info("create new consensus keys")
+			break
+		}
+	}
+
+	// Password is not provided on generating a new consensus keys. Then we need
```
t00ts, line 73: This part of the comment seems a little misleading to me.
What I understood from the code: Password appears to be provided when generating new keys right above. It is *not* provided when the file exists and we have to load it (right below)

pkg/config/samples/default.dusk.toml
```diff
@@ -206,6 +206,9 @@ accumulatorWorkers = 4
 
 # Information for the node to send consensus transactions with
 [consensus]
+# Path to a file that stores Consensus Keys / BLS public and secret keys
+# if file does not exist, it will be created at startup.
+keysfile = "/path/consensus.keys"
```
t00ts, line 211: The default path with `/path/` in it looks weird, but I could be missing something here...

t00ts (approved):
Couple minor comments, but overall looks good

PR #1282, opened 2022-02-04
Parse transaction payload

### LINKED ISSUE #1281: Parse transaction payload structure to populate fields of the Block Explorer
A `decode` method that parses the notes and other useful information from the transaction's payload defined within Rusk

To populate fields in the block explorer that the user needs to see

Augmenting the result values of the gRPC calls `ExecuteStateTransition`, `Accept`, `Verify`

**Additional context**
See dusk-network/rusk#450

herr-seppia:
Reintroduce the old TransactionPayload refactored as TransactionPayloadDecoded.
Add `Decode` to the new TransactionPayload

Resolves #1281


autholykos (changes requested):
LGTM in theory, but please add a test with a real transaction binary

autholykos (approved):
LGTM

PR #1291, opened 2022-02-11
Change state_root to be compatible with rusk

herr-seppia:
Override state_root for tests
Change test_harness to use the downloaded state


autholykos (approved):
LGTM

PR #1295, opened 2022-02-14
Wait txs for being included in the block

### LINKED ISSUE #1293: Stabilize the block time around 15s also for single or a limited Provisioner subset
A `sleep` of sort that will slow down the consensus even when running it with a single Provisioner

The single Provisioner consensus runs at a pace significantly higher than what we expect. Since such a configuration is normally used for testing, this fast pace might lead to tests results inconsistent with what will happen with a configuration with way more nodes. Therefore the consensus time should be normalized

**Additional context**
This was raised by @herr-seppia during his preliminary test-harness results:
```
Block 1 at "time":"2022-02-13T19:38:38Z"
...
Block 24444 at "time":"2022-02-13T21:59:20Z"
```

herr-seppia:
A block is produced only when either:

- there is a transaction in the Mempool
- a 3s timeout commences

Resolves #1293

### DIFF
--- a/pkg/core/consensus/blockgenerator/candidate/blockgenerator.go
+++ b/pkg/core/consensus/blockgenerator/candidate/blockgenerator.go
@@ -136,10 +136,37 @@ func (bg *generator) execute(ctx context.Context, txs []transactions.ContractCal
 	return txs, stateHash, nil
 }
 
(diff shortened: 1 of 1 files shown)


autholykos (approved):
LGTM

Issue #1294, opened 2022-02-14
Assess the real usefulness of the MAGIC header parameter and eventually remove it


autholykos:
**Describe what you want implemented** 
Removal of the MAGIC parameter from the header of the messages

**Describe "Why" this is needed**
The MAGIC has so far been the source of issues and wild goose chases when debugging the protocol. Removing it would probably have no real consequence in the economy of the network

**Describe alternatives you've considered**
Live with it

herr-seppia:
removed in #1337

PR #1297, opened 2022-02-15
Disable magic error propagation

herr-seppia:
This is not going to remove the Magic from the protocol, because is still under review
it just avoid to propagate magic mismatch errors

See also #1294


autholykos (approved):
LGTM

Issue #1299, opened 2022-02-16
Adapt the stake.proto to the latest specification and layout of the Rusk contract


autholykos:
**Describe what you want implemented** 
The protobuf schema should be updated to reflect the changes to the stake so to remove the `extend` and the expiration, as well as adding the `created_at` parameter.

**Describe "Why" this is needed**
Our kids are finally growing and we need to accomodate their new needs

PR #1300, opened 2022-02-16
Adapt stake to new definitions

t00ts:
Resolves: #1299


pkg/core/consensus/user/sortition.go
```diff
@@ -115,8 +115,8 @@ func (p Provisioners) CreateVotingCommittee(seed []byte, round uint64, step uint
 				break
 			}
 
-			if m.Stakes[i].StartHeight > round || m.Stakes[i].EndHeight < round {
-				subtractFromTotalWeight(W, m.Stakes[i].Amount)
+			if m.Stakes[i].CreatedAt > round || m.Stakes[i].Eligibility < round {
```
t00ts, line 118: @autholykos I'm concerned about this, not sure if it's correct anymore with respect to what's now understood as ["valid"](https://github.com/dusk-network/rusk/blob/master/contracts/stake/src/stake.rs#L59-L61)

Would appreciate any insight on the matter...

cc/ @ureeves @goshawk-3

ureeves: Since expiration is now removed, the logic around it should be removed here as well.

`eligibility` is the block at which a stake starts being valid. Assuming that `round` is the current block (unsure about this), the logic should then be exceedingly simple:

```go
if round >= m.Stakes[i].Eligibility {
    // stake is valid here
}
```

pkg/core/consensus/fixtures.go
```diff
@@ -99,8 +99,8 @@ func MockMember(keys key.Keys) *user.Member {
 	}
 
 	member.Stakes = make([]user.Stake, 1)
-	member.Stakes[0].Amount = 500
-	member.Stakes[0].EndHeight = 10000
+	member.Stakes[0].Value = 500
+	member.Stakes[0].Eligibility = 10000
```
herr-seppia, line 103: The eligibility should be `1`, it replace the start_height at line 101

herr-seppia (changes requested):

herr-seppia (commented):
LGTM

herr-seppia (approved):

PR #1312, opened 2022-02-23
Artificially increase block time

herr-seppia:
Increase block time to 15secs
Update rusk state root


pkg/core/consensus/blockgenerator/candidate/blockgenerator.go
```diff
@@ -139,7 +139,7 @@ func (bg *generator) execute(ctx context.Context, txs []transactions.ContractCal
 // fetchOrTimeout will keep trying to FetchMempoolTxs() until either
 // we get some txs or or the timeout expires.
 func (bg *generator) fetchOrTimeout(keys [][]byte) ([]transactions.ContractCall, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
t00ts, line 142: Not a fan of having this `15` hardcoded here

t00ts: Could we add this to `pkg/config/consts.go`?

t00ts (changes requested):

t00ts (approved):

PR #1313, opened 2022-02-24
Recover contract state on node restart

### LINKED ISSUE #1286: The node should make sure that it is in sync with rusk when it restarts
When the node restarts it should check with rusk

Inconsistencies between rusk and the golang process would prevent the node to ever catch up with the blockchain

**Additional context**

goshawk-3:
fixes #1286

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -143,7 +143,72 @@ func New(ctx context.Context, db database.DB, eventBus *eventbus.EventBus, rpcBu
 }
 
 func (c *Chain) syncWithRusk() error {
-	// TODO:
+	ruskStateHash, err := c.proxy.Executor().GetStateRoot(c.ctx)
+	if err != nil {
+		return err
+	}
+
+	prevBlock, persistedHash, err := c.loader.LoadTip()
+	if err != nil {
+		return err
+	}
+
+	// Detect if both services are on the different state
+	var persitedBlock *block.Block
+	err = c.db.View(func(t database.Transaction) error {
+		persitedBlock, err = t.FetchBlock(persistedHash)
+		if err != nil {
+			return err
+		}
+
+		if !bytes.Equal(persitedBlock.Header.StateHash, ruskStateHash) {
+			log.WithField("rusk", hex.EncodeToString(ruskStateHash)).
+				WithField("node", hex.EncodeToString(persitedBlock.Header.StateHash)).
+				Error("invalid state detected")
+			return errors.New("invalid state detected")
+		}
+
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
+	// Update blockchain tip (in-memory)
+	c.tip = persitedBlock
+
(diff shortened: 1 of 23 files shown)


pkg/core/consensus/blockgenerator/candidate/blockgenerator.go
```diff
@@ -139,7 +139,7 @@ func (bg *generator) execute(ctx context.Context, txs []transactions.ContractCal
 // fetchOrTimeout will keep trying to FetchMempoolTxs() until either
 // we get some txs or or the timeout expires.
 func (bg *generator) fetchOrTimeout(keys [][]byte) ([]transactions.ContractCall, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
autholykos, line 142: The timeout for the selection phase was previously around 5-6 seconds. We should not go with 15 seconds for selection as it will lead to very slow block time in a full network

autholykos (changes requested):
I think the timeout for the block generation should be tightened. Other than that LGTM

pkg/core/consensus/blockgenerator/candidate/blockgenerator.go
t00ts, line 142: Not a fan of having this `15` hardcoded here as mentioned in previous PR

goshawk-3: This change came from rebasing with `master`. It's been merged in `master`. Please address this with @herr-seppia 

From master
https://github.com/dusk-network/dusk-blockchain/blob/f9af476da7099f02586507dfaec1d8276072e5c4/pkg/core/consensus/blockgenerator/candidate/blockgenerator.go#L142

pkg/core/chain/chain.go
```diff
@@ -131,20 +129,100 @@ func New(ctx context.Context, db database.DB, eventBus *eventbus.EventBus, rpcBu
...
+	// same then there is no need to execute sync-up.
+	if bytes.Equal(persistedHash, prevBlock.Header.Hash) {
+		return nil
+	}
+
+	// re-accept missing block in order to recover Rusk (unpersisted) state.
+	i := persitedBlock.Header.Height
+
+	for {
+		i++
```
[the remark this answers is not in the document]
goshawk-3: If anyone adds `continue` statement  then `increment it at the end of each loop` will be skipped and it'll end up into infinite loop.

pkg/core/chain/chain.go
```diff
@@ -59,7 +59,7 @@ type Verifier interface {
 // store the blockchain.
 type Loader interface {
 	// LoadTip of the chain.
-	LoadTip() (*block.Block, error)
+	LoadTip() (*block.Block, []byte, error)
```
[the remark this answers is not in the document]
goshawk-3: Valid point. Addressed!

PR #1316, opened 2022-02-28
Check transaction GasLimit against block GasLimit

### LINKED ISSUE #1306: Don't allow transactions equal or higher than block gas limit in the mempool
Transactions that exceed the block gas limit should not be allowed in the mempool.

This prevents transactions that exceed the limit from ever being passed through Rusk.

N/A

**Additional context**
N/A

herr-seppia:
Resolves #1306


ureeves (approved):
I find it odd that we're rejecting transactions that still fit in the block with `>=`, but if this is decision LGTM.

PR #1318, opened 2022-03-01
Test-harness multiple provisioners

### LINKED ISSUE #1317: Add staking smoke tests
Following is a laundry list of checks to be performed to ensure the stability of the network locally:
[ ] lower the maturity period on the stake contract down to 5 blocks
[ ] create 10 stakes
[ ] run a local cluster with 10 Provisioners

The monitoring should focus on memory usage, block time, eventual errors of the consensus, provisioner synchronization

This is needed as a first step toward the full multi-provisioner network cluster.

Create a genesis block with multiple provisioners already embedded. Without lowering the maturity period, this would only enable testing of the consensus, as any stake would require ~20h to become active. Eventually we will have to do both, but not immediately.

**Additional context**

herr-seppia:
- [x] Add reward generator to the rusk command
- [x] ~Send transfer from faucet to provisioner~
- [x] ~Stake from each provisioner~
- [x] Include more stakers in the genesis block

Resolves #1317


pkg/config/consts.go
```diff
@@ -26,7 +26,7 @@ const (
 	MaxInvBlocks = 500
 
 	// Protocol-based consensus step time.
-	ConsensusTimeOut = 20 * time.Second
+	ConsensusTimeOut = 5 * time.Second
```
autholykos, line 29: back to 5s?

herr-seppia: Yes, this was the timeout prior to increase block time to 15secs

pkg/core/data/ipc/transactions/provider.go
```diff
@@ -122,6 +122,8 @@ type executor struct {
 
 // VerifyStateTransition see also Executor.VerifyStateTransition.
 func (e *executor) VerifyStateTransition(ctx context.Context, calls []ContractCall, blockGasLimit, blockHeight uint64) error {
+	time.Sleep(1 * time.Second)
```
autholykos, line 125: What are the sleeps for?

autholykos (commented):
I am fuzzy about the `sleep` before the state transition operations. The rest looks fine

autholykos (approved):
LGTM (sleeps will be tackled in a followup issue)

PR #1323, opened 2022-03-03
Remove Deadline context for Accept, Finalize and Persist rusk grpc calls

### LINKED ISSUE #1200: Disable client-side timeout for grpc ExecuteStateTransition
**Describe the bug**
No timeout grpc at client-side to mitigate any opportunity for inconsistency between dusk and rusk service states.
Dusk node grpc call **ExecuteStateTransition** to Rusk service should never return **DeadlineExceeded** error. On any failure to update Contract storage an error is always returned, respectively no blockchain state update is performed by dusk node.

**Expected behaviour**
Making an assumption that Rusk service will always return an error if it encounters a critical problem, we should threat both services (dusk and rusk) as one.  As a fail-over procedure, we can use `cmd/deployer` executable to restart both services in case of a deadlock situation (Unresponsive service).
 

**Additional context**
https://github.com/dusk-network/dusk-blockchain/pull/1169

goshawk-3:
fixes #1200

### DIFF
--- a/pkg/core/data/ipc/transactions/provider.go
+++ b/pkg/core/data/ipc/transactions/provider.go
@@ -164,9 +164,9 @@ func (e *executor) Finalize(ctx context.Context, calls []ContractCall, stateRoot
 		vstr.Txs[i] = tx
 	}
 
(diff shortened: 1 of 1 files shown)


autholykos (approved):
LGTM

PR #1325, opened 2022-03-04
Introduce Contract Storage revert in Fallback procedure

### LINKED ISSUE #1283: Integrate the state reset when reverting to a finalized block
The RPC client for rusk reset should be implemented and integrated in the `Chain`

Rusk exposes a primitive to reset to the last finalized state. The golang process should use that when detecting forks

No alternatives

**Additional context**
An issue on rusk will follow

### LINKED ISSUE #1326: Add block hash verification in CheckBlockHeader
**Describe the bug**
 `verifiers.CheckBlockHeader` should always ensure that a hashed set of block header fields is equal to the block.header.hash.

goshawk-3:
fixes #1283 
fixes #1326

### DIFF
--- a/pkg/core/data/ipc/transactions/provider.go
+++ b/pkg/core/data/ipc/transactions/provider.go
@@ -46,6 +46,9 @@ type Executor interface {
 
 	// Persist instructs Rusk to persist the state if in-sync with provided stateRoot.
 	Persist(ctx context.Context, stateRoot []byte) error
+
+	// Revert instructs Rusk to revert to the most recent finalized state.
+	Revert(ctx context.Context) error
 }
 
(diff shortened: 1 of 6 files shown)


pkg/core/data/ipc/transactions/provider.go
```diff
@@ -356,6 +359,18 @@ func (e *executor) Persist(ctx context.Context, stateRoot []byte) error {
 	return nil
 }
 
+// Persist proxy call to state.Persist grpc.
```
autholykos, line 362: This comment is wrong. It should be about `Revert`

autholykos (changes requested):
LGTM with a minor comment-related nit

Issue #1240, opened 2022-01-13
Rate limiter should prioritise TXs with higher fees


autholykos:
**Describe what you want implemented** 
A (probably custom made) rate limiter that would prioritize transactions with higher fees

**Describe "Why" this is needed**
Since the introduction of rate limiting, there is an opportunity for propagation of transactions with higher fees first

**Describe alternatives you've considered**
Do nothing and delegate prioritization of transaction execution to the recipient queue. It is actually unclear which approach is the most desirable.

**Additional context**
_Originally posted by @autholykos in https://github.com/dusk-network/dusk-blockchain/pull/1238#discussion_r783784945_

autholykos:
The strategy should also keep into consideration eventual flooding attacks based on burst of invalid transactions. In this case, we should make sure that valid transactions in the Mempool are not dropped nor affected

fed-franz:
Should we close this? @herr-seppia 
We already ARE prioritizing txs based on higher-gas-price strategy.

PR #1330, opened 2022-03-07
Mempool extraction by GasPrice

herr-seppia:
`Fee()` returns GasPrice instead of GasLimit
`Fee()` is used to RangeSort txs extraction

See also dusk-network/rusk#605
See also #1240


goshawk-3 (approved):
LGTM

PR #1332, opened 2022-03-08
Retain transaction status

### LINKED ISSUE #1309: Transaction status should be retained
When a transaction gets executed in the context of a `VerifyStateTransition` call, the execution result should be added to the DB as `transaction status`

A user must be able to check whether its transaction has been executed correctly or its execution failed for any reason. Valid transactions can still fail for different reasons (i.e. _out of gas_, _contract not found_, etc). Such a failure and its causes are information not currently retained anywhere, which causes users to be left in the dark about the operations they perform.

**Additional context**

herr-seppia:
Resolves #1309


autholykos (approved):
LGTM

goshawk-3 (approved):
LGTM

PR #1333, opened 2022-03-09
Throttle block acceptance on specified conditions

### LINKED ISSUE #1331: Slow down consensus agreement time suitably
AcceptBlock procedure should execute for not less than 1s (configurable sleep). The synthetic delay should not be applied in out-sync mode.

This should slow down consensus until more Provisioners join the network or Rusk grpc calls need more than 1s to complete.

FetchOrTimeout in Block generator does the trick too. However it works only if mempool is empty and it does not depend on grpc calls execution time. 

**Additional context**
The additional delay should guarantee a healthy consensus time not less than 5s.

goshawk-3:
fixes #1331

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -12,6 +12,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/dusk-network/dusk-blockchain/pkg/config"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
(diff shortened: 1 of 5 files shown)


pkg/core/chain/chain.go
```diff
@@ -349,6 +372,23 @@ func (c *Chain) acceptSuccessiveBlock(blk block.Block, kadcastHeight byte) error
...
+	// node is in out-of-sync mode.
+	// block time is higher that ConsensusTimeThreshold
+	if prevBlockTimestamp+config.ConsensusTimeThreshold > blk.Header.Timestamp {
+		maxDelayMilli := config.Get().Consensus.ThrottleMilli
+		if maxDelayMilli == 0 {
+			maxDelayMilli = 2000
+		}
+
+		if d, err := throttle(startTime, maxDelayMilli); err == nil {
+			log.WithField("height", blk.Header.Height).WithField("sleep_for", d.String()).Trace("throttled")
```
herr-seppia, line 388: Should `d` be used to wait? Cannot figure out where this duration is used

pkg/core/chain/chain.go
```diff
@@ -336,11 +337,33 @@ func (c *Chain) ProcessSyncTimerExpired(strPeerAddr string) error {
...
+}
+
 // acceptSuccessiveBlock will accept a block which directly follows the chain
 // tip, and advertises it to the node's peers.
 func (c *Chain) acceptSuccessiveBlock(blk block.Block, kadcastHeight byte) error {
+	startTime := time.Now().UnixMilli()
+
 	log.WithField("height", blk.Header.Height).Trace("accepting succeeding block")
 
+	prevBlockTimestamp := c.tip.Header.Timestamp
```
herr-seppia, line 365: I'm a bit concerned to use timestamps in environment where no ntp is configured

Could the received instant be used for this scenario?

pkg/core/chain/chain.go
goshawk-3: Good catch. Forgot to commit it. It's in throttle function.

pkg/core/chain/chain.go
goshawk-3: Timestamp in block header is a field being validated. It's guaranteed that it's bigger than prevBlock timestamp.

herr-seppia: Ok, in this case I open this followUp issue to tackle some odd corner cases

https://github.com/dusk-network/dusk-blockchain/issues/1334

herr-seppia (approved):
LGTM

PR #1337, opened 2022-03-11
Introduce protocol version

### LINKED ISSUE #1319: Protocol versioning behaviour
**The protocol versioning needs to be properly specc'ed**

- If the nodes receive a message with a version that cannot be processed (since it is newer), they should ignore the message
- when we do the handshake with the bootstrapper, the version of the protocol (and the node, and rusk) should be validated. The bootstrapper will probe the version communicated by the nodes and returns a string with the link to where a newer version shall be downloaded
- the ping/pong could be used to communicate/inquiry the protocol version
- in the header of the messages we should add the protocol version
- consensus version shall remain backward compatible (and therefore consensus version should not influence protocol version)
- protocol changes need to be assessed on a case-by-case basis

## Version binary format

The version is represented as a 64-bit number (8 bytes) loosely following[^1] [semantic versioning]( https://semver.org/) with the format below:

| Field | Offset | Type |Description |
| - |  - | - | - |
| RESERVED | 0 | n/a | For future use |
| MAJOR | 2 | `u16` LE[^2] | version with incompatible API changes |
| MINOR | 4 | `u16` LE| version with added functionality in a backwards compatible manner |
| PATCH  | 6 | `u16` LE | version with backwards compatible bug fixes |

### Testnet

Testnet is always considered with `MAJOR` field set to `0`[^3], therefore the semantic of the field is shifted, merging all backwards compatible changes together[^4]:

| Field | Offset | Type |Description |
| - |  - | - | - |
| RESERVED | 0 | n/a | For future use |
| MAJOR | 2 | `u16` LE | Initial development, value blocked to `0x00` |
| MINOR | 4 | `u16` LE | version with incompatible API changes |
| PATCH  | 6 | `u16` LE |version with any backwards compatible changes |

[^1]: We don't have additional labels for pre-release and build metadata, as the semantic versioning allows.
[^2]: Little Endian
[^3]: Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable. https://semver.org/#spec-item-4
[^4]: This is consistent with the semver we have in rust development in Dusk

herr-seppia:
- [x]  Check Protocol version
- [x] Setup Node Version and InterOp Version
- [x] Adapt tests with new txs

Resolves #1319


pkg/config/consts.go
```diff
@@ -35,6 +35,12 @@ const (
 
 	// KadcastInitialHeight sets the default initial height for Kadcast broadcast algorithm.
 	KadcastInitialHeight byte = 128
+
+	// The dusk-blockchain executable version.
+	NodeVersion = "0.5.0"
```
autholykos, line 40: In the future, this should be called `Chain` instead of `Node` to avoid any confusion

pkg/p2p/wire/protocol/frame_test.go
```diff
@@ -25,8 +25,8 @@ func TestWriteReadFrame(t *testing.T) {
 	buf := make([]byte, length)
 	b.Read(buf)
 
-	// Remove magic, checksum and reserved bytes
-	buf = buf[16:]
+	// Remove magic, checksum, version, and reserved bytes
+	buf = buf[(16 + 8):]
```
autholykos, line 29: the magic should be absorbed by the version

pkg/p2p/wire/protocol/protocol.go
```diff
@@ -33,12 +33,22 @@ const (
...
 }
 
+// CurrentProtocolVersion indicates the protocol version used.
+var CurrentProtocolVersion = semver.MustParse("0.1.0")
+
+// VersionConstraintString represent the needed version.
+const VersionConstraintString = "~0.1.0"
+
+// VersionConstraint check incoming versions.
+var VersionConstraint, _ = semver.NewConstraint(VersionConstraintString)
```
autholykos, line 50: Why are these hardcoded and not parsed from configuration?

autholykos (changes requested):
The magic should be absorbed by the version and be stored in one of the reserved bytes. Moreover, I am not sure about hardcoding the version as opposed to storing it in the configuration

autholykos (approved):
LGTM

pkg/config/consts.go
autholykos: This can be the object of a minor followup

PR #1342, opened 2022-03-15
Throttle verifyFn in reduction_1 phase

### LINKED ISSUE #1341: Throttle verifyST to allow agreement loop to propagate AggrAgreement message
While consensus agreement loop is collecting/processing events from iteration N,  it may happen that iteration (N+1) also has been finalized. This is actually a race between agreement process and next iteration execution.

We should try to guarantee that there is enough time for the agreement loop to process events from iteration N. Hence, we mitigate the probabilities of the fallback procedure being needed.

goshawk-3:
fixes #1341


autholykos (approved):
LGTM

PR #1346, opened 2022-03-18
Add staking tx decoding tests

### LINKED ISSUE #1285: Increase tx hash tests to include crossover and calldata
Increase the tests coverage including a binary tx with crossover and calldata

To be sure about the golang hash implementation. Once done we can get rid of rusk calculating the hash for the node

Stick with the current implementation, where the txhash is provided by rusk

**Additional context**
A stake transaction has these two populated

herr-seppia:
Resolves #1285


autholykos (approved):
LGTM

PR #1365, opened 2022-04-06
Call verify block prior to voting for a candidate block

### LINKED ISSUE #1357: Call VerifyStateTransition in Reduction 2 only if it is a committee member
**Describe the bug**
Current implementation calls VST immediately after reaching quorum at reduction 1. However,  VST is supposed to be called prior to voting (as prerequisite) at reduction 2, only if the provisioner is committee member.

**To Reproduce**
Run a cluster with a node that does not  stake. This node will always call VST without participating in consensus ever. 

**Additional context**
This is issue should not be confused with the one for adding VST call in reduction 1.

### LINKED ISSUE #1358: Extend reduction 1 verification procedure with VST call
**Describe the bug**
Current approach:  Any committee member of reduction 1 propagates its vote for a block on two conditions:

- block hash is correctly signed by the Block generator of the iteration
- block header does pass sanity check

In order to be aligned with reduction 2 verification procedure and to enhance security, Red_1 should perform a complete block verification (a.k.a **VST** call)

An **optimization** to be taken into account: If the same provisioner is extracted for both reduction phases of an iteration, it should be enough to call **VST** only once.

goshawk-3:
fixes #1358 
fixes #1357

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -385,7 +385,7 @@ func (c *Chain) acceptSuccessiveBlock(blk block.Block, kadcastHeight byte) error
 			maxDelayMilli = 2000
 		}
 
-		if d, err := util.Throttle(startTime, maxDelayMilli); err == nil {
+		if d, err := util.Delay(startTime, maxDelayMilli); err == nil {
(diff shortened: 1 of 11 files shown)


herr-seppia (approved):
LGTM

PR #1369, opened 2022-04-08
Include BlockGenerator into the header

### LINKED ISSUE #1368: Include GeneratorBlsPubkey in the  block header
The block header should include the GeneratorBlsPubkey and this should be part of the hashed fields.
Pass block.header.GeneratorBlsPubkey into any state transition requests (EST, VST, Accept, Finalize)

This is to increase the clarity of who create the blocks and to better enforce the checks in the distribute phase.
Furthermore, this allow to remove `generator` rusk's CLI args

Stick with the actual implementation

**Additional context**
This could be a good opportunity to get rid of `block.header.TxRoot`
See also #1320

herr-seppia:
Remove `header.TxRoot` and add `header.GeneratorBlsPubkey`

Resolves #1368

### DIFF
--- a/pkg/gql/query/types.go
+++ b/pkg/gql/query/types.go
@@ -53,7 +53,7 @@ var Header = graphql.NewObject(
 			"seed": &graphql.Field{
 				Type: Hex,
 			},
-			"txroot": &graphql.Field{
+			"generatorblspubkey": &graphql.Field{
 				Type: Hex,
 			},
 			"timestamp": &graphql.Field{
--- a/pkg/p2p/wire/encoding/miscdata.go
+++ b/pkg/p2p/wire/encoding/miscdata.go
@@ -107,3 +107,26 @@ func WriteBLS(w *bytes.Buffer, b []byte) error {
 	_, err := w.Write(b)
 	return err
 }
+
+// ReadBLSPKey will read a compressed bls public key (96 bytes) from r and return it as a slice of bytes.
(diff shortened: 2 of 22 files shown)


pkg/gql/query/types.go
```diff
@@ -53,7 +53,7 @@ var Header = graphql.NewObject(
 			"seed": &graphql.Field{
 				Type: Hex,
 			},
-			"txroot": &graphql.Field{
+			"generatorblspubkey": &graphql.Field{
```
goshawk-3, line 56: nitpick: we can also replace "txroot" with "generatorblspubkey" in graphql README as well.

In addition, in `wire.md` doc, we have:

| 32 | Tx Root | \[\]byte | Merkle root hash of all transactions in this block |

pkg/p2p/wire/encoding/miscdata.go
```diff
@@ -107,3 +107,26 @@ func WriteBLS(w *bytes.Buffer, b []byte) error {
...
+	}
+
+	if _, err := r.Read(b); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteBLSPKey will write a compressed bls signature (33 bytes) to w.
```
goshawk-3, line 124: nitpick: update comment

goshawk-3 (approved):
LGTM, just a few minor doc-related issues.

PR #1372, opened 2022-04-12
Decrement kadcast height at Kadcast Writer

### LINKED ISSUE #1362: Decrement kadcast height in kadcast writer
**Describe the bug**
Currently, kadcast height is decremented when kadcast reads a message from the network. However, this would not allow the peer at height 0 to receive the broadcast message.

**To Reproduce**
TBD

goshawk-3:
fixes #1362


herr-seppia (commented):
I merged instead of approve it
But anyway LGTM

We'll refactor this later on (to include the whole metadata structure)

PR #1376, opened 2022-04-15
Skip verifyFn on vote sending if a verified hash is provided

### LINKED ISSUE #1371: Optimize VST calls of a Provisioner per a consensus iteration
As per  https://github.com/dusk-network/dusk-blockchain/pull/1365,  If the same Provisioner is extracted for both reduction phases of a single iteration, it should be enough to call VST only once per block. 

This will reduce the number of VST calls for a Provisioner per an consensus iteration.

goshawk-3:
fixes #1371

### DIFF
--- a/pkg/core/consensus/reduction/reduction.go
+++ b/pkg/core/consensus/reduction/reduction.go
@@ -7,6 +7,7 @@
 package reduction
 
 import (
+	"bytes"
 	"errors"
 	"time"
 
@@ -53,6 +54,8 @@ type Reduction struct {
 
 	// VerifyFn verifies candidate block
 	VerifyFn consensus.CandidateVerificationFunc
+
+	VerifiedHash []byte
 }
 
 // IncreaseTimeout is used when reduction does not reach the quorum or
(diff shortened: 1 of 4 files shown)


pkg/core/consensus/reduction/reduction.go
```diff
@@ -81,6 +85,17 @@ func (r *Reduction) verifyWithDelay(candidate *block.Block, step uint8) ([]byte,
 		return nil, errEmptyBlockHash
 	}
 
+	hash, err := candidate.CalculateHash()
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if we have already verified this block.
+	if len(r.VerifiedHash) == 32 && !bytes.Equal(r.VerifiedHash, block.EmptyHash[:]) && bytes.Equal(r.VerifiedHash, hash) {
```
herr-seppia, line 94: I agree with this. But what if previous step verified a hash that differ from the current step hash? Should the error be raised immediately instead of call verifyFn again?

goshawk-3: In theory, this is an edge case that is allowed to happen but I'll add a warning for now.

herr-seppia (approved):
LGTM

PR #1379, opened 2022-04-26
Verify state hash from VST call

### LINKED ISSUE #1377: The state_root on candidate block verification is not verified
**Describe the bug**
On verifying a candidate block, a provisioner does not check if the state_root calculated in VST call is equal to the one in the candidate block header.  After removing tx_root header field, this problem may allow the network to achieve consensus on a block that will be rejected on `Accept`/ `Finalize` call.

**Expected behaviour**
VST should return the computed state_root on success. In addition, a provisioner should vote for a candidate block only if header state_root is equal to the one returned by VST .

goshawk-3:
fixes #1377

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -37,8 +37,10 @@ import (
 )
 
 var (
-	errInvalidStateHash = errors.New("invalid state hash")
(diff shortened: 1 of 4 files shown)


herr-seppia (approved):
LGTM

PR #1384, opened 2022-05-02
Queue my own vote to be registered locally

### LINKED ISSUE #1360: Register a vote locally before sending my own reduction
**Describe the bug**
Currently, any committee member relies on other provisioners to propagate their own votes back in order to be registered locally. As a side effect of this, `CollectReduction` always need to initiate a kadcast broadcast with `config.KadcastInitHeader`. 

**To Reproduce**
TBD.
 
**Expected behaviour**
As suggested by @herr-seppia , `CollectReduction` should re-propagate kadcast reduction messages instead of initiating a new broadcast (kad_h=128). This to happen, a committee member must register its own first.

goshawk-3:
fixes #1360

### DIFF
--- a/pkg/core/consensus/reduction/firststep/step.go
+++ b/pkg/core/consensus/reduction/firststep/step.go
@@ -11,7 +11,6 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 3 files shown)


herr-seppia (approved):
LGTM

PR #1386, opened 2022-05-03
Discard mempool transactions containing a nullifier already accepted

### LINKED ISSUE #1381: Handle the rejected transactions after a EST
**This issue is a followup of dusk-network/rusk#705**

_Originally posted by @herr-seppia in https://github.com/dusk-network/rusk/pull/705#pullrequestreview-951806041_

goshawk-3:
fixes #1381

### DIFF
--- a/pkg/core/mempool/buntdb.go
+++ b/pkg/core/mempool/buntdb.go
@@ -8,6 +8,7 @@ package mempool
 
 import (
 	"bytes"
+	"errors"
 	"strconv"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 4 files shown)


herr-seppia (changes requested):
I'm ok with this PR.
But I think this address only a part of the linked issue.

It's still missing the part where "all" discarded txs from EST are discarded from the mempool (right now, only the case where a tx is discarded for "usedNullifier" is covered)

herr-seppia (approved):
Approved

Discarding txs right after the EST is dangerous (it's still a candidate block)

PR #1389, opened 2022-05-05
Mempool should discard any transaction with repeated nullifier

### LINKED ISSUE #1388: Mempool should discard any transaction with repeated nullifier
On transaction retrieval, Mempool should check if any of transaction nullifiers is already used by another transaction in mempool state. The check should be executed after a successful rusk.Preverify call.

This change will allow a block generator for passing  transactions to EST without repeated nullifiers. Additionally,  mempool will not need anymore to run a cumbersome `discardUsedNullifiers` at any block acceptance.

goshawk-3:
fixes #1388

### DIFF
--- a/pkg/core/mempool/hashmap.go
+++ b/pkg/core/mempool/hashmap.go
@@ -128,8 +128,8 @@ func (m HashMap) FilterByType(filterType transactions.TxType) []transactions.Con
 	return txs
 }
 
-// Contains returns true if the given key is in the pool.
-func (m *HashMap) Contains(txID []byte) bool {
+// Contain returns true if the given key is in the pool.
+func (m *HashMap) Contain(txID []byte) bool {
 	m.lock.RLock()
 	defer m.lock.RUnlock()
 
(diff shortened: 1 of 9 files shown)


pkg/core/mempool/hashmap.go
```diff
@@ -235,6 +235,37 @@ func (m *HashMap) RangeSort(fn func(k txHash, t TxDesc) (bool, error)) error {
...
+	}
+
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	for _, v := range m.data {
+		d, err := v.tx.Decode()
+		if err != nil {
+			continue
+		}
```
autholykos, line 251: should we really fail silently here? What to do if data is corrupted?

autholykos (approved):
LGTM. Some minor nits that we should probably handle separately

pkg/core/mempool/hashmap.go
goshawk-3: `m.data` represents transactions that pass all checks and are accepted in mempool. If for some odd case `m.data` has accepted a corrupted tx, this (ContainAnyNullifiers) should not fail here, disallowing acceptance of other valid transactions.

PR #1391, opened 2022-05-06
GQL: fix wrong block reward

### LINKED ISSUE #1390: Wrong reward in GQL
**Describe the bug**
Reward shown in GQL endpoint is missing the distribution reward, so it's just the sum of fee paid

**To Reproduce**
GQL query the current devnet
```
  blocks(hash: "a85bcbfbc157ead5922961a6d751c01ef455ee45d90aebfd44dc047e78af5d31") {
    header {
      reward
      feespaid
      height
      timestamp
      step
    }
  }
  ```

**Expected behaviour**
Reward should be the sum of every tx's fee + emission amount 

**Logs/Screenshot**
```json

    "blocks": [
      {
        "header": {
          "feespaid": 723145951,
          "height": 8667,
          "reward": 723145951,
          "step": 3,
          "timestamp": "2022-05-06 10:56:45 +0000 UTC"
        }
      }
    ],
```

**Platform**
N/A

**Additional context**
The reward resolver has not been updated after the removal of the distribute transaction

herr-seppia:
Resolves #1390

### DIFF
--- a/pkg/gql/query/blocks.go
+++ b/pkg/gql/query/blocks.go
@@ -190,6 +190,31 @@ func resolveTxs(p graphql.ResolveParams) (interface{}, error) {
 	return nil, errors.New("invalid source block")
 }
 
+/// This implements the emission schedule described in the economic paper.
+func emissionAmount(blockHeight uint64) uint64 {
+	switch {
+	case blockHeight == 0: // genesis block
+		return uint64(0)
+	case blockHeight <= 12_500_000: // first 12_500_000 blocks emit 16 DUSK
+		return uint64(16 * 1_000_000_000)
+	case blockHeight <= 18_750_000: // from 12_500_001 to 18_750_000 blocks emit 12.8 DUSK
+		return uint64(12.8 * 1_000_000_000)
+	case blockHeight <= 25_000_000: // from 18_750_001 to 25_000_000 blocks emit 9.6 DUSK
+		return uint64(9.6 * 1_000_000_000)
+	case blockHeight <= 31_250_000: // from 25_000_001 to 31_250_000 blocks emit 8 DUSK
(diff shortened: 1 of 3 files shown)


pkg/gql/query/blocks.go
```diff
@@ -199,7 +224,7 @@ func resolveReward(p graphql.ResolveParams) (interface{}, error) {
 			return nil, errors.New("context does not store database conn")
 		}
 
-		reward := uint64(0)
+		reward := emissionAmount(b.Height)
```
goshawk-3, line 227: If this is  the generator reward value, we should subtract the Dusk foundation reward as per `90% of the total value goes to the generator (rounded up)`

pkg/gql/query/blocks.go
```diff
@@ -190,6 +190,31 @@ func resolveTxs(p graphql.ResolveParams) (interface{}, error) {
...
+		return uint64(4.8 * 1_000_000_000)
+	case blockHeight <= 50_000_000: // from 43_750_001 to 50_000_000 blocks emit 3.2 DUSK
+		return uint64(3.2 * 1_000_000_000)
+	case blockHeight <= 62_500_000: // from 50_000_001 to 62_500_000 blocks emit 1.6 DUSK
+		return uint64(1.6 * 1_000_000_000)
+	}
+	return uint64(0) // after 62_500_000 blocks emit 0 DUSK
+}
+
 func resolveReward(p graphql.ResolveParams) (interface{}, error) {
```
goshawk-3, line 218: We should update graphql README to clarify if `reward` field is the total reward value or generator reward only.

goshawk-3 (changes requested):

goshawk-3 (approved):

PR #1392, opened 2022-05-09
Print provisioners list on change. Remove provisioners from StormDB

### LINKED ISSUE #1361: Optimize api.db on cluster
**Describe the bug**
It turns out that api.db is growing fast in size on any cluster node. We should find a way to clean it up on regular base to avoid wasting disk space.

```
 1.2G Apr  5 09:53 /opt/dusk/dusk_data/chain/api.db
```

goshawk-3:
fixes #1361

### DIFF
--- a/pkg/gql/README.md
+++ b/pkg/gql/README.md
@@ -67,18 +67,29 @@ NB: The examples from below represent only query structures. To send a query as 
 {
   blocks(hash: "194dd13ee8a60ac017a82c41c0e2c02498d75f48754351072f392a085d469620" ) {
     header {
-       height
-       hash
-       timestamp
-       version
-       seed
-       prevblockhash
-       statehash
-       generatorblspublickey
-       reward
-       feespaid
+      feespaid            # Total fees paid for all transactions in the block
+      generatorblspubkey  # BLS key of the block generator
+      hash                # Block hash
+      height              # Block height
+      prevblockhash       # Previous block hash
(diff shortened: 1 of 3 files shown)


pkg/core/chain/chain.go
```diff
@@ -489,9 +490,18 @@ func (c *Chain) runStateTransition(tipBlk, blk block.Block) (*block.Block, error
 	// Update the provisioners.
 	// blk.Txs may bring new provisioners to the current state
 	c.p = &provisionersUpdated
+	provNumAfter := c.p.Set.Len()
+
+	// Print provisionres information if provisioner(s) have been
+	// added/removed with this block acceptance
+	if provNumAfter-provNumBefore != 0 || blk.Header.Height == 1 {
```
herr-seppia, line 497: Did you consider the case when there is a provisioner removed and another one added in the same block?
Ps: this can only happen when changing epoch

herr-seppia (approved):
LGTM even if a rare edge case is not covered (but for just logging it maybe it doesn't worth it)

pkg/core/chain/chain.go
goshawk-3: Good point. This could also happen if a provisioner unstakes. Maybe the approach should be revised then. I'll think about it.

goshawk-3: It turns out that we already print provisioners list in consensus if debug level is set, so there is not need to pollute Info log level with duplicated details.

PR #1393, opened 2022-05-09
Disallow provisioners records in StormDB

### LINKED ISSUE #1361: Optimize api.db on cluster

goshawk-3:
fixes #1361


herr-seppia (approved):
LGTM

PR #1397, opened 2022-05-20
Inject rusk's version into gRPC headers

### LINKED ISSUE #1396: Handle rusk grpc Version
Rusk gRPC now requires the expected rusk's version as header. Otherwise it rejects the requests

N/A

**Additional context**
See also dusk-network/rusk#718

herr-seppia:
Resolves #1396

### DIFF
--- a/pkg/core/data/ipc/transactions/provider.go
+++ b/pkg/core/data/ipc/transactions/provider.go
@@ -16,6 +16,7 @@ import (
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/user"
 	"github.com/dusk-network/dusk-protobuf/autogen/go/rusk"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/core/data/ipc/transactions/provider.go
```diff
@@ -392,3 +408,8 @@ func UMember(r *rusk.Provisioner, t *user.Member) {
 		}
 	}
 }
+
+func injectRuskVersion(ctx context.Context) context.Context {
+	md := metadata.New(map[string]string{"x-rusk-version": "0.5.0"})
```
goshawk-3, line 413: Could you move rusk (hard-coded) version  in `pkg/config/consts.go`?

goshawk-3 (changes requested):

pkg/core/data/ipc/transactions/provider.go
herr-seppia: Done

goshawk-3 (approved):
LGTM

PR #1398, opened 2022-05-20
Update mempool state at start-up

### LINKED ISSUE #1258: Sync up mempool state with the network on bootsrapping a node.
**Describe the bug**
In Gossip Network, when a node connects to a peer at startup, it requests a full copy of the Peer's mempool  state via wire message `topics.Mempool`.  

In Kadcast network, a node should be capable of selecting N nodes randomly from the network and requests from them via `NetworkService.Send` method  most recent mempool state.

**To Reproduce**
On (re)starting a node, it will not get most recent transaction from the network.

**Additional context**
Kadcast library should provide a new API to return a set of random addresses of active nodes.

goshawk-3:
fixes #1258


pkg/core/mempool/mempool.go
```diff
@@ -494,6 +491,29 @@ func (m *Mempool) kadcastTx(t TxDesc) error {
...
 
+// RequestUpdates sends topics.MemPool to N Kadcast Network nodes.
+func (m *Mempool) RequestUpdates() {
+	maxNodes := config.
+		Get().
+		Mempool.
+		MaxNumUpdaters
+
+	if maxNodes == 0 {
+		log.Warn("updates disabled")
```
herr-seppia, line 502: Ok for the warning. But I would set a default value greater than 0

pkg/p2p/kadcast/writer/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+	return b
+}
+
+// Subscribe subscribes to eventbus Kadcast messages.
+func (w *Broadcast) Subscribe() {
+	mapper := func(t topics.Topic) byte {
+		const (
+			High = byte(1)
+			Low  = byte(0)
+		)
```
herr-seppia, line 52: Are these priority a new feature? If so a new tracking issue should be created to assess any potential attacking surface

herr-seppia (approved):
LGTM overall

Just a couple of comments to be addressed

pkg/core/mempool/mempool.go
goshawk-3: Valid point.

pkg/p2p/kadcast/writer/broadcast.go
goshawk-3: No, not a new feature. See also master: https://github.com/dusk-network/dusk-blockchain/blob/1d98aa4e1faa4fc1638a8fa8da0471e8c9274cc8/pkg/p2p/kadcast/writer.go#L71

PR #1315, opened 2022-02-25
Fix reward and feesPaid in graphQL

herr-seppia:
Resolves #1314

PR #1417, opened 2022-06-15
Include block gas limit into header

### LINKED ISSUE #1416: Include the used block gas limit into the block header
The blockgaslimit used to produce a block should be decided by the block producer.
In order to let the network validate a block properly, this info should be carried by the block itself so that a verifier can use that limit to validate a block

N/A

**Additional context**
At the moment of writing the block gas limit is hardcoded. This obviously raise a resync problem when a new node join the network after being compiled with a different block gas limit value.
The current hardcoded value should be moved to a configuration value and used as GasLimit only when producing a block.
A further optimisation at verifier side can be to discard any block with gasSpent higher than the gas limit configured

herr-seppia:
Resolves #1416

### DIFF
--- a/pkg/gql/query/types.go
+++ b/pkg/gql/query/types.go
@@ -36,7 +36,10 @@ var Header = graphql.NewObject(
 		Name: "Header",
 		Fields: graphql.Fields{
 			"height": &graphql.Field{
-				Type: graphql.Int,
+				Type: graphql.Float,
+			},
+			"gaslimit": &graphql.Field{
+				Type: graphql.Float,
 			},
 			"hash": &graphql.Field{
 				Type: Hex,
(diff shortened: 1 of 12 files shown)


pkg/gql/query/types.go
```diff
@@ -36,7 +36,10 @@ var Header = graphql.NewObject(
 		Name: "Header",
 		Fields: graphql.Fields{
 			"height": &graphql.Field{
-				Type: graphql.Int,
+				Type: graphql.Float,
```
ureeves, line 39: Is height now a float?

herr-seppia: I forgot to mention that `graphql.Float` is used to map `uint64`, otherwise if the value overflows the `Int.MaxValue` it returns `0`
See also #1315

ureeves (approved):
LGTM

PR #1422, opened 2022-06-20
Improve mempool extraction

### LINKED ISSUE #1421: Improve mempool extraction
Transaction extracted from the mempool should be weighted with the average gas spent for that kind of transaction.
With this weight, the new prioritisation should be `ForecastGasSpent`*`GasPrice`.
Transactions should be extracted until Sum(ForecastGasSpent)>=GasLimit

This has no impact on rusk, if actual Gas is greater than the forecast, transactions will not be included the block (as per the current behaviour)

To extract transactions from the mempool, the only criteria currently used is `GasLimit`*`GasPrice`.
These leave the stage to two unwanted scenario:
1. With higher GasLimit the prioritisation of a transfer transaction can be higher than a stake transaction, even if the latter will effectively use more gas 
2. All transactions that fit the BlockSize limit are executed to try to fit into a block (even if there is no reasonable gas left). This highly impact ExecuteStateTransition performance and, therefore, the whole consensus performance

N/A

**Additional context**
See also #1240

herr-seppia:
Resolves #1421


pkg/core/data/ipc/transactions/payload_decoded.go
```diff
@@ -193,3 +193,17 @@ func UnmarshalTransactionPayloadDecoded(r *bytes.Buffer, f *TransactionPayloadDe
...
+// is an intercontract call.
+// TODO: further improvement should take care of the contractAddress to compare
+// it against historical data.
+func (p *TransactionPayloadDecoded) EstimatedGasSpent() uint64 {
+	if p.Call == nil {
+		// This is a normal transfer transaction
+		return 250_000_000
+	}
+	// This is an inter-contract call
+	return 1_200_000_000
```
t00ts, line 208: Would suggest doing it the other way around.
The "normal" transaction is what should be in the bottom-most return.
The "non-normal" (inter-contract call) should be in the `if` block return.

```go
	if p.Call != nil {
		// This is an inter-contract call
		return 1_200_000_000
	}
	// Normal transfer transactions
	return 250_000_000
```

pkg/core/mempool/mempool.go
```diff
@@ -447,16 +448,42 @@ func (m Mempool) processGetMempoolTxsBySizeRequest(r rpcbus.Request) (interface{
 	}
 
 	txs := make([]transactions.ContractCall, 0)
+	gasLimit := config.Get().State.BlockGasLimit
+	// The slippageGasLimit is the threshold that consider the "estimated gas
+	// spent" acceptable even if it exceeds the strict GasLimit. This is
+	// required to avoid to iterate the whole meempol until it fit perfectly
+	// the block GasLimit
+	slippageGasLimit := gasLimit + gasLimit/10
```
t00ts, line 456: Might as well do:
```go
slippageGasLimit := gasLimit * 1.10
```

pkg/core/mempool/mempool.go
```diff
@@ -447,16 +448,42 @@ func (m Mempool) processGetMempoolTxsBySizeRequest(r rpcbus.Request) (interface{
...
+		// If the totalGas exceeded the gasLimit (but still below the
+		// slippage threshold) stop to iterate the mempool
+		done = totalGas >= gasLimit
+
 		totalSize += uint32(t.size)
 
 		if totalSize <= maxTxsSize {
 			txs = append(txs, t.tx)
 		} else {
+			// it the totalSize exceeds the limit, stop to iterate the mempool
```
t00ts, line 486: Maybe this last bit could just be

```go
		if totalSize <= maxTxsSize {
			txs = append(txs, t.tx)
		}

		// We're done when the totalGas exceeds the gasLimit or
		// totalSize exceeds the tx size limit
		done = totalGas >= gasLimit || totalSize > maxTxsSize
```

t00ts (approved):
LGTM, with couple nits.

pkg/core/data/ipc/transactions/payload_decoded.go
herr-seppia: In the future the "inter-contract call" branch will be longer (as per historical comparison)

So I prefer to leave it outer

pkg/core/mempool/mempool.go
herr-seppia: not exactly 

If we exceed the size, we must not include the tx
If we exceed the gas limit (but not the slippage) we have to include it

This is why I've split the checks

pkg/core/mempool/mempool.go
t00ts: Ah, missed the types... 🤷‍♂️ ✅

pkg/core/mempool/mempool.go
herr-seppia: Refactored right now to improve readability

PR #1424, opened 2022-06-21
Fix consensus timeout configuration

### LINKED ISSUE #1423: Make consensus timeout configurable
To help the tuning phase during stress tests and benchmarking

N/A

**Additional context**
N/A

herr-seppia:
ConsensusTimeout is now initialized with DefaultConsensusTimeout (5 seconds) but can be overrode using the custom configuration

Resolves #1423

### DIFF
--- a/pkg/config/consts.go
+++ b/pkg/config/consts.go
@@ -6,8 +6,6 @@
 
 package config
 
-import "time"
-
 // A single point of constants definition.
 const (
 	// DUSK is one whole unit of DUSK.
@@ -40,7 +38,7 @@ const (
 
 	// Consensus-related settings
 	// Protocol-based consensus step time.
-	ConsensusTimeOut = 5 * time.Second
+	DefaultConsensusTimeOutSeconds = 5
 
 	// ConsensusTimeThreshold consensus time in seconds above which we don't throttle it.
 	ConsensusTimeThreshold = 10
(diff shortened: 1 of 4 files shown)


pkg/config/loader.go
```diff
@@ -256,7 +256,7 @@ func init() {
 	r.Wallet.Store = "walletDB"
 	r.Consensus.DefaultLockTime = 1000
 	r.Consensus.DefaultAmount = 10
-	r.Consensus.ConsensusTimeOut = 5
+	r.Consensus.ConsensusTimeOut = DefaultConsensusTimeOutSeconds
```
t00ts, line 259: Unrelated to this PR, but as per Go's recommended naming schemes, `ConsensusTimeout` should just be `Timeout` in this context, as it already belongs to `Consensus` category.

pkg/config/consts.go
```diff
@@ -40,7 +38,7 @@ const (
 
 	// Consensus-related settings
 	// Protocol-based consensus step time.
-	ConsensusTimeOut = 5 * time.Second
+	DefaultConsensusTimeOutSeconds = 5
```
t00ts, line 41: Pure curiosity: Can we have the const value be a `time.Duration`? Something that may be worth considering.

t00ts (approved):
LGTM!

pkg/config/loader.go
herr-seppia: I agree, but this PR aimed to be backward compatible with the previous configuration files

PR #1426, opened 2022-06-28
Detect if a received block originated from a fork

### LINKED ISSUE #1412: Trace counts of both eligible and non-eligible provisioners
**Describe the bug**
`Prov` logging field currently shows the length of all network provisioners. It should also show the number of eligible provisioners for the round of the accepted block .

### LINKED ISSUE #1413: Detect any occurrences of missed fallback procedure
**Describe the bug**
In a situation of a consensus split, the subset of the network of the wrong branch may reach quorum and accept another block. The edge case should be detected in runtime and all details logged.

**To Reproduce**
1. Enforce a race condition between agreement and consensus loop
2. Delay the propagation of the block from the lowest iteration (e.g block.cert_step = 3)
3. As a result network should manage to continue with the branch with block.cert_step = 6 and accept another block.
4. Propagate the block with cert_step = 3

**Logs/Screenshot**
Blockchain Fallback - Use Case 3(1)

### LINKED ISSUE #1425: Drop topics.Block messages with invalid header.hash value
**Describe the bug**
A block received from the network should not be processed if block.Header.Hash is invalid.

**To Reproduce**
1. Marshal a block with  Header.Hash different from `block.CalculateHash()`
2. Broadcast the block as  `topics.Block` message

**Expected behaviour**
Block should be rejected in `ProcessBlockFromNetwork` as this is already done by the handler of `topics.Candidate`

goshawk-3:
fixes #1413 
fixes #1425 
fixes #1412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search txs up to 10.000 blocks [#1401]
 - Mempool updates its state at start-up automatically [#1258]
 - Mempool discards any transaction with repeated nullifier [#1388]
+- Detect any occurrences of missed fallback procedure [#1413] 
 
 ### Changed
 - Extraction from mempool consider Gas expenditure estimation [#1421]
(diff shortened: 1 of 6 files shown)


herr-seppia (approved):
LGTM

PR #1427, opened 2022-06-30
Request candidate block from  arbitrary active nodes

### LINKED ISSUE #1359: Improve  GetCandidate request mechanism
**Describe the bug**
Current implementation. In an edge case, a provisioner may discard the iteration candidate block. As a result, this provisioner will rely on the `GetCandidate` request mechanism to fetch it from the entire network. 

GetCandidate` request mechanism does propagate a request to the entire network and waits a few seconds for a valid respond. This mechanism introduces an inefficiency in consensus and different option could be discussed to improve it.

goshawk-3:
fixes #1359  (Implements Option_1)


herr-seppia (approved):
LGTM

PR #1431, opened 2022-07-05
Check certificate of the winning block

### LINKED ISSUE #1430: Discard aggrAgreement messages with invalid StepVotes
**Describe the bug**
Any wire message should be fully verified. Currently, StepVotes from  aggrAgreement are fully trusted without any checks.

**To Reproduce**
Option_1: Send aggrAgreement with StepVotes with total votes less than the quorum target.
Option_2: Send aggrAgreement with StepVotes invalid signature.

**Expected behaviour**
we should always ensure the  aggrAgreement from the wire is fully valid before we create the winning block and certificate.

### LINKED ISSUE #1432: Check quorum target on certificate verification
**Describe the bug**
Currently, quorum target is checked only on receiving an agreement. However Chain component needs to ensure the certificate of the winning block includes StepVotes with enough votes (target quorum of the round)

goshawk-3:
fixes #1430 
fixes #1432

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/dusk-network/dusk-blockchain/pkg/config"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus"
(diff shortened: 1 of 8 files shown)


herr-seppia (approved):
LGTM

PR #1434, opened 2022-07-06
Use eventbus.NewSafeChanListener on subscribing for topics.Agreement

### LINKED ISSUE #1433: Clone message.Agreement on notifying the consumer
**Describe the bug**
2th_Reduction sends a message.Agreement internally with `StepVotes` references of both 1th and 2th Reduction. The message is collected by another goroutine (Agreement loop) and processed concurrently in `handler.Verify`.  That creates a DATA RACE situation - while Consensus loop is changing `firstStepsVotes`, the same stepVotes (Signature and bitsets) are being read by Agreement loop.

**To Reproduce**
```go
       	ev := message.MockAgreement(hash, round, step, keys, p)
	handler := NewHandler(keys[0], *p, []byte{0, 0, 0})


	a.CreateWorkers(workersCount)

	for i := 0; i < msgCount; i++ {
		a.Process(ev)
	}

	ev.VotesPerStep[0].BitSet *= 2 // DATA RACE
```
As a result , we get `error="failed to verify BLS multisig: verification failed"`

**Expected behaviour**
We should pass a deep copy of message.Agreement to Agreement loop.

**Additional context**
This is probably the culprit of `"failed to verify BLS multisig: verification failed` spotted on `devnet`

goshawk-3:
fixes #1433

### DIFF
--- a/pkg/core/loop/consensus.go
+++ b/pkg/core/loop/consensus.go
@@ -88,7 +88,7 @@ func New(e *consensus.Emitter) *Consensus {
 	aggrAgreementChan := make(chan message.Message, msgChanSize)
 
 	// subscribe agreement phase to message.Agreement
-	aChan := eventbus.NewChanListener(agreementChan)
+	aChan := eventbus.NewSafeChanListener(agreementChan)
 	e.EventBus.Subscribe(topics.Agreement, aChan)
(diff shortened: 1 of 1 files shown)


herr-seppia (approved):
LGTM

PR #1437, opened 2022-07-08
Allow verifyFn to report errors correctly

### LINKED ISSUE #1436: Allow verifyFn to report errors correctly
**Describe the bug**
If a node receives next valid block and accepts it (state updated)  while consensus goroutine is still running next iteration for the previous round then `1st_reduction/2th_reduction` steps  will fail on `verifyFn` as per the logs below. This pollutes logs with misleading errors which should be avoided gracefully.

**Expected behaviour**
`verifyFn` errors should alarm for a invalid candidate blocks. 

**Logs/Screenshot**
```
verifyfn failed - previous block hash does not equal the previous hash in the current block
verifyfn failed - block already exists
```

**Additional context**
Noticed on devnet but should be reproducible on test-harness as well.

goshawk-3:
fixes #1436 

NB Tested on devnet - verifyFn errors are reported only on essential problems.


herr-seppia (approved):
LGTM

PR #1442, opened 2022-07-14
Send reduction concurrently without blocking votes re-propagation

### LINKED ISSUE #1441: Investigate consensus delay when mempool is full
We are experiencing block time (**40-50 seconds**) above the average (8seconds throttled) when the mempool is full.
Atm, the block gas limit is set to ~3B, this means that we have 9-10 transfers per block.
In this case, EST calls took up to 4 seconds, and VST/Accept calls took 2-3 secs

Assuming we have different provisioners for each step the following time should be spent in execute/wait StateTransitions:
* Selection - wait for the block produced with EST: 4seconds.
* 1st Reduction - VST: 3 seconds.
* 2nd Reduction - VST: 3 seconds.  
* Accepting a block (after Agreement) - Accept/Finalize: 3 seconds

**Total: 13 seconds**

To these, we should add the network propagation (not relevant for now).

**Additional context**
Enabling the auto propagation flag on kadcast doesn't change the behaviour

goshawk-3:
fixes #1441 

Additionally, another edge case will be addressed out of this PR (with issue https://github.com/dusk-network/dusk-blockchain/issues/1444)

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -32,6 +32,7 @@ import (
 	"github.com/dusk-network/dusk-blockchain/pkg/util/diagnostics"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/eventbus"
 	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/rpcbus"
+	"github.com/dusk-network/dusk-blockchain/pkg/util/nativeutils/sortedset"
 	"github.com/dusk-network/dusk-protobuf/autogen/go/node"
 	"github.com/sirupsen/logrus"
 	logger "github.com/sirupsen/logrus"
@@ -111,6 +112,7 @@ type Chain struct {
 	ctx context.Context
 
 	blacklisted dupemap.TmpMap
+	verified    sortedset.SafeSet
 }
 
 // New returns a new chain object. It accepts the EventBus (for messages coming
@@ -129,6 +131,7 @@ func New(ctx context.Context, db database.DB, eventBus *eventbus.EventBus, rpcBu
 		loop:              loop,
 		stopConsensusChan: make(chan struct{}),
 		blacklisted:       *dupemap.NewTmpMap(1000, 120),
+		verified:          sortedset.NewSafeSet(),
 	}
 
 	chain.synchronizer = newSynchronizer(db, chain)
@@ -615,6 +618,7 @@ func (c *Chain) acceptBlock(blk block.Block, withSanityCheck bool) error {
 	}
 
 	c.tip = b
+	c.verified.Reset()
 
 	// 5. Perform all post-events on accepting a block
 	c.postAcceptBlock(*b, l)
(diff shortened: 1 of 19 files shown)


pkg/core/consensus/reduction/firststep/step.go
```diff
@@ -98,10 +98,15 @@ func (p *Phase) Run(ctx context.Context, queue *consensus.Queue, evChan chan mes
...
-		r.Round, step, &p.selectionResult.Candidate)
-	// }
+	a := reduction.NewAsyncSend(p.Reduction, r.Round, step, &p.selectionResult.Candidate)
+
+	if p.handler.AmMember(r.Round, step) {
+		_ = a.Go(ctx, &wg, evChan, reduction.Republish)
+	} else {
+		if p.handler.AmMember(r.Round, step+1) {
+			_ = a.Go(ctx, &wg, evChan, reduction.ValidateOnly)
+		}
```
herr-seppia, line 108: I like this trick

pkg/core/consensus/reduction/firststep/step.go
```diff
@@ -94,14 +94,21 @@ func (p *Phase) Run(ctx context.Context, queue *consensus.Queue, evChan chan mes
...
 
 	p.handler = reduction.NewHandler(p.Keys, r.P, r.Seed)
-	// first we send our own Selection
-	if p.handler.AmMember(r.Round, step) {
-		m, _ := p.SendReduction(r.Round, step, &p.selectionResult.Candidate)
 
-		// Queue my own vote to be registered locally
-		evChan <- m
+	// send our own Selection
+	var wg sync.WaitGroup
```
herr-seppia, line 99: Any reason to use a `WaitGroup` without waiting for it?

pkg/core/consensus/reduction/secondstep/step.go
```diff
@@ -96,11 +96,15 @@ func (p *Phase) Run(ctx context.Context, queue *consensus.Queue, evChan chan mes
 	p.handler = reduction.NewHandler(p.Keys, r.P, r.Seed)
 	// first we send our own Selection
 
-	if p.handler.AmMember(r.Round, step) {
-		m, _ := p.SendReduction(r.Round, step, p.firstStepVotesMsg.Candidate)
+	// wait group
+	var wg sync.WaitGroup
```
herr-seppia, line 100: Any reason to use a WaitGroup without waiting for it?

pkg/core/chain/chain.go
```diff
@@ -384,11 +386,13 @@ func (c *Chain) ProcessSyncTimerExpired(strPeerAddr string) error {
 // acceptSuccessiveBlock will accept a block which directly follows the chain
 // tip, and advertises it to the node's peers.
 func (c *Chain) acceptSuccessiveBlock(blk block.Block, kadcastHeight byte) error {
-	startTime := time.Now().UnixMilli()
-
 	log.WithField("height", blk.Header.Height).Trace("accepting succeeding block")
 
-	prevBlockTimestamp := c.tip.Header.Timestamp
+	// TODO: Verify Certificate
```
herr-seppia, line 391: Is this a leftover? If not, a follow up issue would fit the case

pkg/core/consensus/reduction/firststep/step.go
```diff
@@ -94,14 +94,21 @@ func (p *Phase) Run(ctx context.Context, queue *consensus.Queue, evChan chan mes
...
+
+	a := reduction.NewAsyncSend(p.Reduction, r.Round, step, &p.selectionResult.Candidate)
+
+	if p.handler.AmMember(r.Round, step) {
+		_ = a.Go(ctx, &wg, reductionChan, reduction.Republish)
```
herr-seppia, line 104: Shall the returned cancel function be deferred?

herr-seppia (commented):
Not sure about some code snippets

pkg/core/consensus/reduction/firststep/step.go
goshawk-3: In 1th_reduction step, It is not deferred on purpose.

pkg/core/chain/chain.go
goshawk-3: Good catch. I'll file it

pkg/core/consensus/reduction/firststep/step.go
goshawk-3: It's part of the input params for  AsyncSend.Go, but not using it anymore. So, I'll remove them.

pkg/core/chain/chain.go
goshawk-3: FYI https://github.com/dusk-network/dusk-blockchain/issues/1446

herr-seppia (approved):
LGTM

PR #1445, opened 2022-07-27
Reintroduce artificial delay on block generator

herr-seppia:


pkg/config/groups.go
```diff
@@ -137,6 +137,8 @@ type mempoolConfiguration struct {
 
 	// Enables mempool updates at startup
 	Updates updates
+
+	ExtractionDelaySecs uint32
```
goshawk-3, line 141: nitpick. Add comment here to explain the new field.

goshawk-3 (changes requested):
Maybe it's also good to benchmark the PR with empty and full blocks in devnet to ensure we have a predictable result.

goshawk-3 (approved):

herr-seppia:
> @goshawk-3 
> Maybe it's also good to benchmark the PR with empty and full blocks in devnet to ensure we have a predictable result.

Prior to this PR:

image
image
image
image

With this PR
image
image
image
image

goshawk-3:
Looks very good @herr-seppia

PR #1447, opened 2022-08-01
Verify block header before repropagating

### LINKED ISSUE #1446: Verify block certificate prior to broadcast
**Describe the bug**
On receiving a block from Kadcast network, we used to fully verify it before re-propagating it to the next bucket. Now to speed up block broadcast to all provisioners, we re-propagate it without any checks. 

Checking a certificate and block header should be a precondition for a block propagation.

goshawk-3:
fix #1446


herr-seppia (approved):
LGTM

PR #1458, opened 2022-08-26
Use FillBytes in Provisioners.MemberAt

### LINKED ISSUE #1457: Add support in sortition for  bls public keys that starts with zeros
**Describe the bug**
` p.MemberAt(i)` in **extractCommitteeMember** would return `nil` member making the sortition panic, when a provisioner  public key starts with zeros.

**To Reproduce**
1. Try to generate a valid BLS key where the public key starts with zeros
2. Stake with it
3. If extractCommitteeMember loop reaches this member, it will panic.

**Expected behaviour**
As it is an existing item, p.MemberAt(i) should return the real member but not nil. To achieve this, `Int.FillBytes` should be used.

**Additional context**
```golang
func TestSetBytes(t *testing.T) {
	b1 := decodeString("0000d081884c")
	bigInt1 := new(big.Int).SetBytes(b1)

	b2 := bigInt1.Bytes()

	assert.True(t, bytes.Equal(b1, b2))
	// failed
}
```

goshawk-3:
fixes #1457


pkg/core/consensus/user/provisioners.go
```diff
@@ -120,7 +123,7 @@ func NewProvisioners() *Provisioners {
 
 // Add a Member to the Provisioners by using the bytes of a BLS public key.
 func (p *Provisioners) Add(pubKeyBLS []byte, value, reward, counter, eligibility uint64) error {
-	if len(pubKeyBLS) != 96 {
+	if len(pubKeyBLS) != BlsKeySize {
 		return fmt.Errorf("public key is %v bytes long instead of 96", len(pubKeyBLS))
```
herr-seppia, line 127: Better to change the hardcoded `96` in the message

pkg/core/consensus/user/provisioners.go
```diff
@@ -196,8 +199,8 @@ func (p Provisioners) GetMember(pubKeyBLS []byte) *Member {
 // GetStake will find a certain provisioner in the committee by BLS public key,
 // and return their stake.
 func (p Provisioners) GetStake(pubKeyBLS []byte) (uint64, error) {
-	if len(pubKeyBLS) != 96 {
-		return 0, fmt.Errorf("public key is %v bytes long instead of 129", len(pubKeyBLS))
+	if len(pubKeyBLS) != BlsKeySize {
+		return 0, fmt.Errorf("public key is %v bytes long instead of 96", len(pubKeyBLS))
```
herr-seppia, line 203: Same as previous

herr-seppia (changes requested):
LGTM overall

RFC due to missing updates in the CHANGELOG

herr-seppia (approved):
LGTM

PR #1459, opened 2022-08-26
Remove deprecated committee keys from block generator

### LINKED ISSUE #1456: Remove deprecated committee keys from block generator
**Describe the bug**
`regenerateCommittee`  is called in `GenerateCandidateMessage` but its return value is never used.

goshawk-3:
fixes #1456


herr-seppia (approved):
LGTM

Issue #1448, opened 2022-08-03
Prevent "ghost" provisioners


herr-seppia:
**Describe what you want implemented** 
A mechanism to prevent "ghost" provisioners to keep participating in the consensus.

**Describe "Why" this is needed**
If a provisioner is not active (both in good and bad faith) it can significantly impact the consensus performance when it's selected as block generator
To address this issue, we need to find a way to prevent (or at least lower the probability for) a "ghost" node to be selected.

**Proposed solution**
Since the block generator selection is deterministic, when a block is not accepted at the first iteration the network knows exactly which node (bls key) failed to produce an accepted block.
If a node keep continuing to not produce block over the time (let's think in term of epochs) it can be safely detected by the whole network. 
In this case, a slashing mechanism should be triggered and included in the block. 

This slashing mechanism can be:
1. proportional to the amount of missed blocks
2. percentage fixed but increased in case of missing consecutive epochs (eg: 20% - 40% - 70% - 100%) 3. 

To prevent slashing honest nodes that missed to produce block for a spot reason, a threshold should be taken into consideration.
E.g: node need to miss more than 30% of blocks (with at least 5 block missed)

Since this is not been incepted to harm provisioners, the slashed amount should not be removed from circulation (as other chains are used to do) but should just removed from the staked amount
> ⚠️ This mechanism can be exploited by user who want to withdraw their stake without paying fee (they can simply power off their nodes and wait for their stake amount to be slashed). To prevent this, the slashed amount should remain in the stake-contract state to be withdrawn as usual

**Describe alternatives you've considered**
Incentivise those provisioners who partecipate properly in the consensus. 
Btw this:
1. don't prevent "ghost" provisioners to be selected, but it just lower (a little) their probability increasing the other nodes probability 
4. exclude incentivization to the provisioners that are already cutoff by the quorum mechanism (only 67% will be rewarded)

**Additional context**
In order to prevent malicious nodes to play the game, staking withdraw should be allowed only after a period of unlocking

PR #1471, opened 2022-11-04
Add log for missed block detection

### LINKED ISSUE #1470: Missing block detection
Logging of missed and produced blocks in base58, along with the public key belonging to the privisioner.

It would be great if node operators would be able to figure out if a certain provisioner missed a block during the consensus round. We can use this as a way to calculate "uptime" in terms of participation. 

It can help both the operator to figure out if something is wrong at their end but also the wider network to spot weaker nodes.

**Additional context**
It's not possible to figure out uptime for nodes in the network directly or to ping them and deduce uptime from that. 

This logging approach should give us close to what we want, without exposing new APIs or sacrificing node privacy/load.

herr-seppia:
See also #1448 
Resolves #1470

### DIFF
--- a/pkg/core/chain/chain.go
+++ b/pkg/core/chain/chain.go
@@ -19,6 +19,7 @@ import (
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/capi"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/reduction"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/consensus/user"
+	"github.com/dusk-network/dusk-blockchain/pkg/core/data/base58"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/data/block"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/data/ipc/transactions"
 	"github.com/dusk-network/dusk-blockchain/pkg/core/database"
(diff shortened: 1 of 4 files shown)


pkg/core/chain/chain.go
```diff
@@ -509,6 +521,12 @@ func (c *Chain) runStateTransition(tipBlk, blk block.Block) (*block.Block, error
 		WithField("state_hash", util.StringifyBytes(respStateHash)).WithField("e_prov", eligibleProvisioners).
 		Info("state transition completed")
 
+	provisioner, _ := base58.Encode(blk.Header.GeneratorBlsPubkey)
+	logger.WithField("provisioner", provisioner).
```
goshawk-3, line 525: nitpick: Should we name it  block_generator or just "generator" instead of "provisioner"?

pkg/core/chain/chain.go
```diff
@@ -459,6 +460,17 @@ func (c *Chain) runStateTransition(tipBlk, blk block.Block) (*block.Block, error
 			return block.NewBlock(), err
 		}
 	default:
+		missedIterations := blk.Header.Step/3 - 1
+		for iteration := uint8(0); iteration < missedIterations; iteration++ {
+			step := iteration*3 + 1
+			committee := c.p.CreateVotingCommittee(tipBlk.Header.Seed, blk.Header.Height, step, config.ConsensusSelectionMaxCommitteeSize)
+			expectedkey, _ := base58.Encode(committee.MemberKeys()[0])
```
goshawk-3, line 467: As a rule of thumb, always ensure "committee.MemberKeys()" is not empty.

goshawk-3 (changes requested):

goshawk-3 (approved):

PR #1480, opened 2022-12-05
Fetch provisioners list from stake contract only if the list is updated

### LINKED ISSUE #1479: Fetch provisioners list from stake contract only if the list is updated
**Describe the bug**
`GetProvisioners` call consumes more than 4s (up to 15s on low-spec nodes) allegedly due to serialization in VM. Although this is not supposed to be the case with the optimized VM (next version),  this should be handled gracefully from client-side (Golang part). 

**To Reproduce**
1. Introduce 2000-size allow list
2. Call Accept/Finalize (as a result this calls GetProvisioners).

**Expected behaviour**
Until next VM is benchmark-ed and fully integrated, dusk node should not call `grpc.GetProvisioners` if provisioners list is not updated.

goshawk-3:
fixes #1479

### DIFF
--- a/pkg/core/data/ipc/transactions/provider.go
+++ b/pkg/core/data/ipc/transactions/provider.go
@@ -36,10 +36,10 @@ type Executor interface {
 	ExecuteStateTransition(context.Context, []ContractCall, uint64, uint64, []byte) ([]ContractCall, []byte, error)
 
 	// Accept creates an ephemeral state transition.
-	Accept(context.Context, []ContractCall, []byte, uint64, uint64, []byte) ([]ContractCall, user.Provisioners, []byte, error)
+	Accept(context.Context, []ContractCall, []byte, uint64, uint64, []byte, *user.Provisioners) ([]ContractCall, user.Provisioners, []byte, error)
 
 	// Finalize creates a finalized state transition.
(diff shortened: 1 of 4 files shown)


pkg/core/data/ipc/transactions/provider.go
```diff
@@ -190,7 +190,17 @@ func (e *executor) Finalize(ctx context.Context, calls []ContractCall, stateRoot
 		return nil, user.Provisioners{}, nil, err
 	}
 
+	resCalls, err := e.convertToContractCall(res.Txs)
+	if err != nil {
+		return nil, user.Provisioners{}, nil, err
```
herr-seppia, line 195: If we fail to convert txs we should get a fresh provisioners list instead of propagating errors.

This will never happen with the current implementation, but seems safer to me

pkg/core/data/ipc/transactions/provider.go
```diff
@@ -244,6 +275,15 @@ func (e *executor) Accept(ctx context.Context, calls []ContractCall, stateRoot [
...
```
herr-seppia, line 280: Ditto

pkg/core/data/ipc/transactions/provider.go
goshawk-3: Maybe let's handle this with another issue/PR.  In this PR, I'm just moving the convertToContractCall call a few lines above - before GetProvisioners.

herr-seppia (approved):
LGTM

PR #1482, opened 2022-12-21
Add `raw` field to the transaction

### LINKED ISSUE #1481: GraphQL - expose raw transaction data
Add a graphQL field to return the raw transaction binary 

This is required by the wallet to get the full transaction history

herr-seppia:
Resolves #1481

### DIFF
--- a/pkg/gql/query/transactions.go
+++ b/pkg/gql/query/transactions.go
@@ -54,6 +54,7 @@ type (
 		JSON           string
 		TxError        string
 		ContractInfo   *contractInfo `json:"contractinfo"`
+		Raw            []byte
 	}
 )
 
@@ -80,6 +81,7 @@ func newQueryTx(tx core.ContractCall, blockHash []byte, timestamp int64, blockhe
 	qd.GasPrice = decoded.Fee.GasPrice
 
 	qd.GasSpent = tx.GasSpent()
+	qd.Raw = tx.StandardTx().Data
 
 	qd.BlockHash = blockHash
 	qd.BlockTimestamp = timestamp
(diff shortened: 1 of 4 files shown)


pkg/gql/query/types.go
```diff
@@ -171,6 +175,36 @@ var Hex = graphql.NewScalar(graphql.ScalarConfig{
...
+			return b64.StdEncoding.EncodeToString(value)
+		default:
+			return nil
+		}
+	},
+	// ParseValue parses GraphQL variables from `string` to `[]byte`.
+	ParseValue: func(value interface{}) interface{} {
+		switch value := value.(type) {
+		case string:
+			bytes, _ := b64.StdEncoding.DecodeString(value)
```
goshawk-3, line 195: If we can't return the err, maybe we can just log it as an error.

pkg/gql/query/transactions.go
```diff
@@ -80,6 +81,7 @@ func newQueryTx(tx core.ContractCall, blockHash []byte, timestamp int64, blockhe
 	qd.GasPrice = decoded.Fee.GasPrice
 
 	qd.GasSpent = tx.GasSpent()
+	qd.Raw = tx.StandardTx().Data
```
goshawk-3, line 84: Could you update the gql README with an example usage of the new field?

goshawk-3 (approved):
LGTM, just two nitpicks.

pkg/gql/query/types.go
herr-seppia: Removed as we don't need to parse it at all

pkg/gql/query/transactions.go
herr-seppia: Added to the example with the full supported fields list

PR #1501, opened 2023-02-21
Add `Iteration` to block header

### LINKED ISSUE #1500: Include `step` into certificate header
The `step` should be included into the hashable fields used to generate the block hash

To avoid proposing blocks for different iteration with the same block hash when the proposer is the same

N/A

**Additional context**
N/A

herr-seppia:
This add `iteration` to the block header, including it into the hashed fields used to calculate the block hash

In addition, this PR remove `step` from block header certificate (since it's a duplicated info)

Resolves #1500


goshawk-3 (approved):
LGTM

PR #1512, opened 2023-03-22
Fix createSortitionHash description

### LINKED ISSUE #1509: Fix createSortitionHash description
**Issue**
In `createSortitionHash`, the `i` parameter is indicated as _iteration number_, which sounds mistakenly related to the SA iterations. 

**Solution**
Fix description stating that  `i` is actually the number of the committee member being extracted.

fed-franz:
Resolves #1509

### DIFF
--- a/pkg/core/consensus/user/sortition.go
+++ b/pkg/core/consensus/user/sortition.go
@@ -77,8 +77,9 @@ func (v VotingCommittee) MarshalJSON() ([]byte, error) {
 	return json.Marshal(data)
 }
 
-// createSortitionHash takes Seed value 'seed', round number 'round', step number 'step', and iteration number 'i',
-// and returns the SHA3-256 hash of their concatenation (i.e., H(round||i||step||seed)).
+// createSortitionHash takes the Seed value 'seed', the round number 'round', the step number 'step',
+// the index 'i' of the committee member to extract,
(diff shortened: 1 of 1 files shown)


pkg/core/consensus/user/sortition.go
```diff
@@ -77,7 +77,7 @@ func (v VotingCommittee) MarshalJSON() ([]byte, error) {
 	return json.Marshal(data)
 }
 
-// createSortitionHash takes Seed value 'seed', round number 'round', step number 'step', and iteration number 'i',
+// createSortitionHash takes Seed value 'seed', round number 'round', step number 'step', and committee member number 'i',
```
herr-seppia, line 80: What about index `instead` of `number`?
```suggestion
// createSortitionHash takes Seed value 'seed', round number 'round', step number 'step', and committee member index 'i',
```

fed-franz: I wasn't convinced by index, because it can be confused with the index of the member in the provisioner list,
but I guess conceptually it's correct to consider 'i' as the index in the committee that has to be "filled in" with a new member (i.e. the one we're extracting now).

Maybe we could say it explicitly. Like:
"..., and the index 'i' of the committee member to be extracted." ?

herr-seppia: Yes, let's go for the explicitness

herr-seppia (approved):
LGTM

PR #1518, opened 2023-04-05
Use config.DUSK in sortition

### LINKED ISSUE #1507: Update README on Deterministic Sortition
After #1491, 
summarize Deterministic Sortition algorithm in [this](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/core/consensus/user/README.md) README.

### LINKED ISSUE #1517: DUSK value is wrong
**ISSUE**
In [`sortition.go`](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/core/consensus/user/sortition.go#L25), 
DUSK (1 unit of DUSK) is set to be 100 000 000 instead of 1 000 000 000.

**SOLUTION**
Set DUSK = 1000000000
or
use the definition in [`consts.go`](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/config/consts.go#L12)

fed-franz:
Resolves #1517

### DIFF
--- a/pkg/core/consensus/user/sortition.go
+++ b/pkg/core/consensus/user/sortition.go
@@ -13,16 +13,12 @@ import (
 	"math"
 	"math/big"
 
+	"github.com/dusk-network/dusk-blockchain/pkg/config"
 	"github.com/dusk-network/dusk-blockchain/pkg/util"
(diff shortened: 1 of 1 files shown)


herr-seppia (approved):
LGTM

Pls submit the issue on the rust implementation repo too (I expect it to have the same error since we were able to reach a quorum)

PR #1524, opened 2023-04-20
Fix extractCommitteeMember loop

### LINKED ISSUE #1511: Fix extractCommitteeMember
**QUESTIONS on [`extractCommitteeMember`](https://github.com/dusk-network/dusk-blockchain/blob/master/pkg/core/consensus/user/sortition.go#L168)**
_Originally posted by @fed-franz in https://github.com/dusk-network/dusk-blockchain/issues/1491#issuecomment-1419435084_

Original code:
```
	for i := 0; ; i++ {
		// If a provisioner is missing, we use the provisioner at position 0
		if m, e = p.MemberAt(i); e != nil {
			m, e = p.MemberAt(0)

			// If provisioner 0 is also missing, panic
			if e != nil {
				// FIXME: shall this panic?
				log.Panic(e)
			}

			i = 0
		}

		stake, err := p.GetStake(m.PublicKeyBLS)
		if err != nil {
			// If we get an error from GetStake, it means we either got a public key of a
			// provisioner who is no longer in the set, or we got a malformed public key.
			// We can't repair our committee on the fly, so we have to panic.
			log.Panic(fmt.Errorf("pk: %s err: %v", util.StringifyBytes(m.PublicKeyBLS), err))
		}

		// If the current stake is higher than the score, return the current provisioner's BLS key
		if stake >= score {
			return m.PublicKeyBLS
		}

		score -= stake
	}
}
```

This line:
`if m, e = p.MemberAt(i); e != nil {`
handles the case in which a Member is missing.
Q: When does/can this occur?

This line:
`m, e = p.MemberAt(0)`
handles a missing Member by setting the current member 'm' to Member 0.
Q: What's the logic behind this? 

This line:
`if err != nil {`
seems similar to the case in which a member is missing. 
Q: When does this occur? What's the difference between the two cases?

fed-franz:
Resolves #1511

### DIFF
--- a/pkg/core/consensus/user/README.md
+++ b/pkg/core/consensus/user/README.md
@@ -63,11 +63,11 @@ Parameters:
 
 Procedure:
 ```
-i = 0                                           // Start with provisioner at index 0
+i = 0                                           // Start with the first provisioner in the list
 loop:
     1. provisioner = Provisioners[i]            // Get provisioner at index `i`
     2. if provisioner.stake >= score            // If the provisioner's stake is higher than `score`,
         return provisioner                      // extract the provisioner
     4. score = score - provisioner.stake        // Otherwise, decrement the score by the provisioner's stake
-    5. i++                                      // And move to the next provisioner
+    5. i++ % Provisioners.size()                // And move to the next provisioner (loop over when reaching the last index)
 ```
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

PR #1527, opened 2023-05-03
Use constant Committee Size

### LINKED ISSUE #1520: Make CommitteeSize consistent with size
**Issue**
The `CommitteeSize` function decides the committee size depending on the number of provisioners.
This is somewhat inconsistent with the committee 'size' as number of voting credits.

In particular, given how this function behaves, the committee size is set to less than `ConsensusMaxCommitteeSize` (64) if the number of provisioners is lower than this value.

**Solution**
We should decide if this behavior is correct and whether we want it or not.

**Notes**
It is important to note that this function is also used to compute the `quorum` for  the consensus.
That is, the number of votes to reach the quorum is computed using `CommitteeSize`

fed-franz:
This PR removes the CommitteeSize() function, which computed the committee `size` as the minimum between MaxCommitteeSize and the number of available provisioners.
As a result, it makes all functions simply use the constant MaxCommitteeSize as the committee size.

The PR also fixes several tests, whose behavior wrongly assumed a correlation between the number of provisioners and the committee size.

Resolves #1520

### DIFF
--- a/pkg/config/consts.go
+++ b/pkg/config/consts.go
@@ -43,12 +43,12 @@ const (
 	// ConsensusMaxStep consensus max step number.
 	ConsensusMaxStep = uint8(213)
 
-	// ConsensusCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize is the (constant) size of the committee in
 	// 1st_Reduction, 2th_Reduction and Agreement phases.
 	ConsensusCommitteeSize = 64
 
-	// ConsensusCommitteeSize represents the maximum size of the committee in
-	// Selection phase.
+	// ConsensusSelectionCommitteeSize is the size of the committee in
+	// the Selection phase.
 	ConsensusSelectionCommitteeSize = 1
 
 	// RuskVersion is the version of the supported rusk binary.
--- a/pkg/config/consts.go
+++ b/pkg/config/consts.go
@@ -43,13 +43,13 @@ const (
 	// ConsensusMaxStep consensus max step number.
 	ConsensusMaxStep = uint8(213)
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize represents the maximum size of the committee in
 	// 1st_Reduction, 2th_Reduction and Agreement phases.
-	ConsensusMaxCommitteeSize = 64
+	ConsensusCommitteeSize = 64
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize represents the maximum size of the committee in
 	// Selection phase.
-	ConsensusSelectionMaxCommitteeSize = 1
+	ConsensusSelectionCommitteeSize = 1
 
 	// RuskVersion is the version of the supported rusk binary.
 	RuskVersion = "0.6.0"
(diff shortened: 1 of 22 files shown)


herr-seppia (approved):
LGTM

But I strongly suggest to wait for @goshawk-3 feedback too

pkg/core/consensus/user/sortition.go
```diff
@@ -150,7 +150,7 @@ func (p Provisioners) CreateVotingCommittee(seed []byte, round uint64, step uint
 
 		// Deduct up to 1 DUSK from the extracted member's stake.
 		m := p.GetMember(blsPk)
-		subtracted := m.SubtractFromStake(1 * config.DUSK)
+		subtracted := m.SubtractFromStake(10)
```
goshawk-3, line 153: What is the meaning of value 10 here? The comment  says `deduct up to 1 DUSK`.

pkg/config/consts.go
```diff
@@ -43,13 +43,13 @@ const (
...
 	ConsensusMaxStep = uint8(213)
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize represents the maximum size of the committee in
 	// 1st_Reduction, 2th_Reduction and Agreement phases.
-	ConsensusMaxCommitteeSize = 64
+	ConsensusCommitteeSize = 64
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize represents the maximum size of the committee in
```
goshawk-3, line 50: This should be `ConsensusSelectionCommitteeSize` in the comment (seems like an old typo).

Also remove `maximum size of `.

pkg/config/consts.go
```diff
@@ -43,13 +43,13 @@ const (
 	// ConsensusMaxStep consensus max step number.
 	ConsensusMaxStep = uint8(213)
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
+	// ConsensusCommitteeSize represents the maximum size of the committee in
```
goshawk-3, line 46: Replace  `maximum size of ` with `constant size of`.

pkg/config/consts.go
```diff
@@ -43,13 +43,13 @@ const (
 	// ConsensusMaxStep consensus max step number.
 	ConsensusMaxStep = uint8(213)
 
-	// ConsensusMaxCommitteeSize represents the maximum size of the committee in
```
goshawk-3, line 46: As you're renaming this, why don't you use the new terminology -  `number of credits/number of slots` ?

pkg/core/consensus/user/sortition.go
goshawk-3: If I'm not mistaken, we substract `1_000_000_000` in the rust implementation?
FYI @herr-seppia

goshawk-3 (changes requested):

pkg/core/consensus/user/sortition.go
herr-seppia: You're right, good catch

This is a leftover of some tests

pkg/config/consts.go
fed-franz: I was planning to do the renaming globally with a separate PR

pkg/core/consensus/user/sortition.go
fed-franz: Restored to 1 Dusk

PR #1542, opened 2025-03-03
fix: fix the wrong error return value

drawdrop:
In fact, err is incorrect and should be e.


fed-franz:
This repo is deprecated. Please refer to https://github.com/dusk-network/rusk for the current protocol implementation

